PR #14, opened 2020-07-01
Eliminate array copy in NormalizeOp

am15h:


xunkai55:
@googlebot rescan

xunkai55:
Though it's not really "ready to pull", I'd assign this label to trigger an internal code sync, thus I can benchmark the change. Please stay tuned. Thanks!

xunkai55:
Per benchmarking, the change has negative impact on performance of a followed Quantization Op in a typical image preprocessing pipeline:

Resize - Crop - Rotate - Normalize - Quantize

And performance of "Normalization" Op is not improved.

At the first glance, I suspect there's something with ByteBuffer.

Let's hold this for a while and I'll try spare some time investigating it.

PR #586, opened 2021-07-08
Create BUILD_INSTRUCTIONS_FOR_LINUX.md

milindthakur177:
Build Instructions for Linux

### DIFF
--- a/BUILD_INSTRUCTIONS_FOR_LINUX.md
+++ b/BUILD_INSTRUCTIONS_FOR_LINUX.md
@@ -1,0 +1,39 @@
+# Build Instructions for Linux
+
+## Required Libraries
+*   Python2
+*   TensorFlow
+*   Git
+*   Bazel-3.7.2
+
+
+## Installation Instructions
+
+#### For Python2
+In the console, run:
+```bash
+sudo apt install python
+```
+
+#### For TensorFlow
+In the console, run:
+```bash
+sudo apt install python3-pip
+pip install tensorflow
+```
+
+#### For Git
+In the console, run:
+```bash
+sudo apt-get update
+sudo apt-get install git
+```
+
+#### For Bazel
+Download bazel binary installer from [Bazel release page on GitHub](https://github.com/bazelbuild/bazel/releases/tag/3.7.2). 
+In the console, run:
+```bash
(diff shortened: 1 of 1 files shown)


milindthakur177:
@googlebot I signed it!

BUILD_INSTRUCTIONS_FOR_LINUX.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+#### For Git
+In the console, run:
+```bash
+sudo apt-get update
+sudo apt-get install git
+```
+
+#### For Bazel
```
xunkai55, line 32: Thanks for the contribution, @milindthakur177 !

I would recommend using bazelisk here rather than installing Bazel directly. Could you help try that and update?

BUILD_INSTRUCTIONS_FOR_LINUX.md
```diff
@@ -0,0 +1,39 @@
+# Build Instructions for Linux
+
+## Required Libraries
+*   Python2
```
xunkai55, line 4: In theory we shouldn't depend on Python2. Could you please help point out where we require Py2?

BUILD_INSTRUCTIONS_FOR_LINUX.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Installation Instructions
+
+#### For Python2
+In the console, run:
+```bash
+sudo apt install python
+```
+
+#### For TensorFlow
```
xunkai55, line 18: Let's just skip the python thing here, as developers usually have their own way to manage python (pyenv, virtualenv, conda, ...)

xunkai55 (changes requested):

BUILD_INSTRUCTIONS_FOR_LINUX.md
milindthakur177: Sure, I'll try bazelisk and update the instructions. I installed bazel directly since the installation instructions are quite straightforward.

BUILD_INSTRUCTIONS_FOR_LINUX.md
milindthakur177: You are right python version does not matter. I'll update it as well.

PR #612, opened 2021-07-21
Support for dynamic input shape

jonpsy:
This should, to some extent, solve  tensorflow/tensorflow#50448. @lu-wang-g  Some help here would be great.


tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc
```diff
@@ -178,9 +178,32 @@ StatusOr<ImageTensorSpecs> BuildInputImageTensorSpecs(
...
+                                       TfLiteSupportStatus::kInvalidArgumentError);
+      
+      const int height_mutable = dims_signature->data[1];
+      const int width_mutable = dims_signature->data[2];
+
+      if (height_mutable || width_mutable)
+      {
+        // How do I get the image height and width???
+        height = height_mutable ? height : ???;
+        width = width_mutable ? width : ???;
```
jonpsy, line 201: @lu-wang-g  @khanhlvg  some help here?

jonpsy, line 201: Since it is dynamic, how do I get image height and width? Do I need additional args in ```BuildInputTensorSpecs``` ? @lu-wang-g @khanhlvg

jonpsy: We could get image height and width from the ```const FrameBuffer& frame_buffer``` arg, but then ```BuildInputTensorSpecs``` doesn't have access to it. Should I move this code to ```Preprocess()```?

tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc
```diff
@@ -177,13 +177,32 @@ StatusOr<ImageTensorSpecs> BuildInputImageTensorSpecs(
...
+  {
+      if (dims_signature->data[3] == -1 || dims_signature->data[0] == -1)
+        return CreateStatusWithPayload(StatusCode::kInvalidArgument,
+                                       "Only height and width is mutable for now.",
+                                       TfLiteSupportStatus::kInvalidArgumentError);
+      
+      const int height_mutable = dims_signature->data[1];
+      const int width_mutable = dims_signature->data[2];
+      
+      // How do I get the image height and width???
```
lu-wang-g: Height and width won't be populated here, because FrameBuffer is passed in through [BaseVisionTaskApi::Preprocess](https://github.com/tensorflow/tflite-support/blob/ba16516f0f06558a38201958253b42e14c913447/tensorflow_lite_support/cc/task/vision/core/base_vision_task_api.h#L115) or [ImagePreprocessor::Preprocess](https://github.com/tensorflow/tflite-support/blob/f52f2d9b121aae7754eb512006a8734c113d7175/tensorflow_lite_support/cc/task/processor/image_preprocessor.h#L76). 

After a second thought, let's make changes directly to ImageProcessor instead of BaseVisionTaskApi, so that we don't need to move the changes from BaseVisionTaskApi to ImageProcessor later. What do you think?

tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc
lu-wang-g: Didn't see your questions here first. But yes, I think you should move the code to Preprocess() instead.

tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc
jonpsy: Okay but does it have access to the interpreter (to resize the graph and to access input tensor)?

lu-wang-g: TfLiteEngine is a member of the base classes, Preprocessor and Postprocessor. See the code here: https://github.com/tensorflow/tflite-support/blob/f52f2d9b121aae7754eb512006a8734c113d7175/tensorflow_lite_support/cc/task/processor/processor.h#L106

jonpsy: Ah , the engine was the wrapper to the interpreter if I recall correctly. So i could use ```engine_``` for resize operator. Correct?

lu-wang-g: That's right. You can retrieve Interpreter from tflite_engine.

jonpsy:
Made the changes. @lu-wang-g  let me know if I need to clarify my stance on any part of the code.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
```diff
@@ -91,7 +110,7 @@ absl::Status ImagePreprocessor::Preprocess(const FrameBuffer& frame_buffer,
   std::unique_ptr<FrameBuffer> preprocessed_frame_buffer;
   std::vector<uint8> preprocessed_data;
 
-  if (IsImagePreprocessingNeeded(frame_buffer, roi)) {
+  if (IsImagePreprocessingNeeded(frame_buffer, roi) && !IsInputImageDynamic()) {
```
lu-wang-g, line 113: What is an image needs preprocessing (cropping, rotation) while the input shape is dynamic?

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
```diff
@@ -114,6 +133,38 @@ absl::Status ImagePreprocessor::Preprocess(const FrameBuffer& frame_buffer,
   } else {
     // Input frame buffer already targets model requirements: skip image
     // preprocessing. For RGB, the data is always stored in a single plane.
+
+    // If dynamic, it will re-dim the entire graph as per the input.
+    if (IsInputImageDynamic()) {
```
lu-wang-g, line 138: We can compute IsInputImageDynamic in Init and save it to a bool, same as height_mutable and width_mutable.

jonpsy: We would need to create a member to store this state then. Is that okay?

jonpsy:
Should we allow mutability for batch and colour params? I could throw an error if their ```dims signature``` is -1. Something like this

```c++
    // Only the HxW dimensions support mutability.
    const bool batch_and_color_fixed = dims_signature->data[0] != -1 &&
                                       dims_signature->data[3] != -1;
    
    if (!batch_and_color_fixed)
      return tflite::support::CreateStatusWithPayload(
          absl::StatusCode::kUnimplemented,
          "Only Height and Width are mutable for now.");
```

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
lu-wang-g: Of course. Feel free to add extra member fields.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
jonpsy: For rotation, I think for all outputs kTopLeft is fixed right? So for ```preprocesed_frame```,  I need to set input_specs_.width = roi.width similarly for height. The orientation can remain the same (kTopLeft).

lu-wang-g:
> Should we allow mutability for batch and colour params? I could throw an error if their `dims signature` is -1. Something like this
> 
> ```c++
>     // Only the HxW dimensions support mutability.
>     if (dims_signature->data[0] == -1 || dims_signature->data[3] == -1)
>       return tflite::support::CreateStatusWithPayload(
>           absl::StatusCode::kUnimplemented,
>           "Only Height and Width are mutable for now.");
> ```

Yes, the batch axis should always be 1 and depth should always be 3. See the code [here](https://github.com/tensorflow/tflite-support/blob/ba16516f0f06558a38201958253b42e14c913447/tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc#L190).

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
lu-wang-g: output_buffer kTopLeft is fixed right, but you'll need to reorient the output_buffer to the same orientation of the input image. See the [example of image_segmenter](https://github.com/tensorflow/tflite-support/blob/ba16516f0f06558a38201958253b42e14c913447/tensorflow_lite_support/cc/task/vision/image_segmenter.cc#L396-L397).

I need to set input_specs_.width = roi.width similarly for height => Sounds good!

jonpsy:
@lu-wang-g When I use ```bazel test``` command the ```image_classifier``` test passes but when I run the executable in ```bazel-bin``` folder some tests fail.

jonpsy:
We need a dynamic input TFLite model for testing. A perfect candidate would be [this](https://github.com/tensorflow/tensorflow/files/6712927/dilated_conv.zip) model from tensorflow/tensorflow#50448.  A tentative list of things we can test:

a) ```TEST_F(DynamicInput, OutputTensorShape)```: Test if the output tensor shape is dynamic as per the given input image.
b) ```TEST_F(DynamicInput, ErrorOnBatchOrDepthChange)``` : Throw error if ```batch``` or ```depth``` dim signature is -1.
c) ```TEST_F(DynamicInput, PreprocessSucess)```: With a reference input  + processed image, see if the output image is similar to the post-processed output tensor. We need to write the ```PostProcess``` logic for it first. I guess we can delay this one till that happens.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
```diff
@@ -68,6 +68,22 @@ absl::Status ImagePreprocessor::Init(
         "ImagePreprocessor only supports RGB color space for now.");
   }
 
+  // Determine if the input shape is resizable.
+  const TfLiteIntArray* dims_signature = Tensor()->dims_signature;
+
+  if (dims_signature != nullptr)
```
lu-wang-g, line 74: Can you please format your code into the google style? See the [answer](https://stackoverflow.com/a/23408687) here about how to use clang-format.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
```diff
@@ -68,6 +68,22 @@ absl::Status ImagePreprocessor::Init(
...
+  {
+    const bool height_or_width_mutable = dims_signature->data[1] == -1 || 
+                                         dims_signature->data[2] == -1;
```
lu-wang-g, line 77: Nit: merge this line with line 85, such as:

is_input_dynamic_ = dims_signature->data[1] == -1 ||  dims_signature->data[2] == -1;

lu-wang-g: Also, consider to save is_height_mutable_ / is_width_mutable_ instead of is_input_dynamic_.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
```diff
@@ -91,10 +107,20 @@ absl::Status ImagePreprocessor::Preprocess(const FrameBuffer& frame_buffer,
   std::unique_ptr<FrameBuffer> preprocessed_frame_buffer;
   std::vector<uint8> preprocessed_data;
 
+  // Determine if the input shape is resizable.
+  const TfLiteIntArray* dims_signature = Tensor()->dims_signature;
+  const bool height_mutable = dims_signature->data[1] == -1;
+  const bool width_mutable = dims_signature->data[2] == -1;
```
lu-wang-g, line 113: No need to add const before bool.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
```diff
@@ -119,6 +145,29 @@ absl::Status ImagePreprocessor::Preprocess(const FrameBuffer& frame_buffer,
...
+    const int expected_height = Tensor()->dims->data[1];
+    const int expected_width = Tensor()->dims->data[2];
+    const int expected_depth = Tensor()->dims->data[3];
+
+    // HxW of the input image.
+    const int image_height = roi.height();
+    const int image_width = roi.width();
+
+    const int height = height_mutable ? image_height : expected_height;
+    const int width = width_mutable ? image_width : expected_width;
```
lu-wang-g, line 164: Just use input_specs_.image_width / input_specs_.image_height for width / height?

Again, no need to add const in front.

tensorflow_lite_support/cc/test/task/processor/BUILD
```diff
@@ -0,0 +1,45 @@
...
+)
+
+cc_test_with_tflite(
+    name = "image_preprocessor_test",
+    srcs = ["image_preprocessor_test.cc"],
+    data = [
+        "//tensorflow_lite_support/cc/test/testdata/task/vision:test_images",
+        "//tensorflow_lite_support/cc/test/testdata/task/vision:test_models",
+    ],
+    tflite_deps = [
```
lu-wang-g, line 20: A lot of the  tflite_deps and deps can be removed.

If a target is cc_library_with_tflite, use tflite_deps, otherwise, if it is cc_library, use deps. cc_library_with_tflite has not difference with cc_library in the github version, but Google internally, they will pull in different tflite version.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
jonpsy: That's ingenious, thanks!

jonpsy: About the const, I'm doing it to establish that the values won't change. imo, it will improve readability and help the compiler as well.

jonpsy:
Thanks for the suggestions, the code looks much cleaner now!

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
jonpsy: As discussed in meeting, kTopLeft is set for ```output_buffer```. Resolving this.

lu-wang-g:
> We need a dynamic input TFLite model for testing. A perfect candidate would be [this](https://github.com/tensorflow/tensorflow/files/6712927/dilated_conv.zip) model from [tensorflow/tensorflow#50448](https://github.com/tensorflow/tensorflow/issues/50448). A tentative list of things we can test:

Can you please ask the author and see if he agrees that we can use his model for testing? Otherwise, I'll need to search internally and  see if there's any model can be open sourced.

> 
> a) `TEST_F(DynamicInput, OutputTensorShape)`: Test if the output tensor shape is dynamic as per the given input image.
> b) `TEST_F(DynamicInput, ErrorOnBatchOrDepthChange)` : Throw error if `batch` or `depth` dim signature is -1.
> c) `TEST_F(DynamicInput, PreprocessSucess)`: With a reference input + processed image, see if the output image is similar to the post-processed output tensor. We need to write the `PostProcess` logic for it first. I guess we can delay this one till that happens.

The above plan sounds good to me. Thanks for sharing it with me Sai!

jonpsy:
The author is fine with using his model.

jonpsy:
@lu-wang-g  I want to load a model and an image. I think image_preprocessor has a separate engine for itself, but I'm not sure how I can feed data into it and extract data. Thanks in advance!

lu-wang-g:
Do you mean by load model and image in the unit test? Some unit tests (especially those check the results of classify) in image_classifier_test.cc are not open sourced yet. Here is code example of ImageClassifier for your reference:
```
StatusOr<ImageData> LoadImage(std::string image_name) {
  return DecodeImageFromFile(JoinPath(absl::GetFlag(FLAGS_test_srcdir),
                                      kTestDataDirectory, image_name));
}

TEST(ClassifyTest, SucceedsWithFloatModel) {
  ASSERT_OK_AND_ASSIGN(ImageData rgb_image, LoadImage("burger.jpg"));
  std::unique_ptr<FrameBuffer> frame_buffer = CreateFromRgbRawBuffer(
      rgb_image.pixel_data,
      FrameBuffer::Dimension{rgb_image.width, rgb_image.height});

  ImageClassifierOptions options;
  options.set_max_results(3);
  options.mutable_model_file_with_metadata()->set_file_name(
      JoinPath(absl::GetFlag(FLAGS_test_srcdir), kTestDataDirectory,
               kMobileNetFloatWithMetadata));

  ASSERT_OK_AND_ASSIGN(std::unique_ptr<ImageClassifier> image_classifier,
                       ImageClassifier::CreateFromOptions(options));

  StatusOr<ClassificationResult> result_or =
      image_classifier->Classify(*frame_buffer);
  ImageDataFree(&rgb_image);
  ASSERT_OK(result_or);

  const ClassificationResult& result = result_or.value();
  EXPECT_THAT(
      result,
      Approximately(EqualsProto(
          R"pb(classifications {
                 classes {
                   index: 934
                   score: 0.7399742
                   class_name: "cheeseburger"
                 }
                 classes {
                   index: 925
                   score: 0.026928535
                   class_name: "guacamole"
                 }
                 classes { index: 932 score: 0.025737215 class_name: "bagel" }
                 head_index: 0
               }
          )pb")));
}
```

All the header files and dependencies are available in the open sourced version of image_classifier_test.cc. The only reason that the above test is not open sourced is because Approximately and EqualsProto are not available in the OSS version of googletest. We'll need to fork it or reimplement it. But for your test, Approximately and EqualsProto are not needed, because you only cares about comparing images.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
```diff
@@ -68,6 +68,14 @@ absl::Status ImagePreprocessor::Init(
...
   }
 
+  // Determine if the input shape is resizable.
+  const TfLiteIntArray* dims_signature = Tensor()->dims_signature;
+
+  if (dims_signature != nullptr) {
+    // Only the HxW dimensions support mutability.
+    is_height_mutable_ = dims_signature->data[1] == -1;
+    is_width_mutable_ = dims_signature->data[2] == -1;
+  }
```
jonpsy, line 78: @lu-wang-g  I was thinking, we can allow batch, depth to be mutable. If, however, the values of batch and depth are invalid (not 1, 3 respectively) it will be caught in **BuildInputSpecs()** method. What's your thoughts?

jonpsy:
The current problem is that, input tensor is being re-sized correctly but the output tensor shape remains the same. How do I reshape the output tensor? I checked `interpreter.h` and it doesn't have `ResizeTensor()`, even though it's implemented in `subgraph.cc`

jonpsy:
> The current problem is that, input tensor is being re-sized correctly but the output tensor shape remains the same. How do I reshape the output tensor? I checked `interpreter.h` and it doesn't have `ResizeTensor()`, even though it's implemented in `subgraph.cc`

Thanks to the answer from the issue I created, I just had to call `Invoke()` method. All tests pass! :partying_face:

jonpsy:
Unable to get absolute path for ```std::ifstream```. :confused:

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
lu-wang-g: But BuildInputImageTensorSpecs() is only invoked in Init(). It won't catch it If batch and depth changed in ImagePreprocessor::Preprocess(), right?

tensorflow_lite_support/cc/test/task/processor/BUILD
```diff
@@ -0,0 +1,34 @@
...
+cc_test_with_tflite(
+    name = "image_preprocessor_test",
+    srcs = ["image_preprocessor_test.cc"],
+    data = [
+        "//tensorflow_lite_support/cc/test/testdata/task/vision:test_images",
+        "//tensorflow_lite_support/cc/test/testdata/task/vision:test_models",
+    ],
+    deps = [
+        "@org_tensorflow//tensorflow/lite/core/shims:cc_shims_test_util",
+        "//tensorflow_lite_support/cc/task/core:task_api_factory",
```
lu-wang-g, line 22: task_api_factory can be removed, as well as its header file in image_preprocessor_test.cc.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
```diff
@@ -0,0 +1,147 @@
...
+
+using ::testing::ElementsAreArray;
+using ::testing::HasSubstr;
+using ::testing::Optional;
+using ::tflite::support::kTfLiteSupportPayload;
+using ::tflite::support::StatusOr;
+using ::tflite::support::TfLiteSupportStatus;
+using ::tflite::task::JoinPath;
+using ::tflite::task::core::PopulateTensor;
+using ::tflite::task::core::TaskAPIFactory;
```
lu-wang-g, line 50: PopulateTensor and TaskAPIFactory can be removed.

tensorflow_lite_support/cc/test/task/processor/BUILD
```diff
@@ -0,0 +1,34 @@
...
+        "//tensorflow_lite_support/cc/task/processor:image_preprocessor",
+        "//tensorflow_lite_support/cc/port:gtest_main",
+        "//tensorflow_lite_support/cc/task/vision/utils:frame_buffer_common_utils",
+        "//tensorflow_lite_support/cc/test:test_utils",
+        "//tensorflow_lite_support/examples/task/vision/desktop/utils:image_utils",
+        "@com_google_absl//absl/flags:flag",
+        "@com_google_absl//absl/strings:cord",
+        "@org_tensorflow//tensorflow/lite:framework",
+        "@org_tensorflow//tensorflow/lite/c:common",
+        "@org_tensorflow//tensorflow/lite/kernels:builtin_ops",
```
lu-wang-g, line 32: line 29 - 32 are not needed either.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
```diff
@@ -0,0 +1,147 @@
...
+using ::tflite::task::core::TfLiteEngine;
+using ::tflite::task::vision::DecodeImageFromFile;
+using ::tflite::task::vision::FrameBuffer;
+using ::tflite::task::vision::ImageData;
+
+constexpr char kTestDataDirectory[] =
+    "tensorflow_lite_support/cc/test/testdata/task/vision/";
+
+constexpr char kDilatedConvolutionModelWithMetaData[] = "dilated_conv.tflite";
```
lu-wang-g, line 59: Does this model has metadata? Otherwise remove `WithMetaData` from the name.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+ protected:
+  std::unique_ptr<ImagePreprocessor> preprocessor_ = nullptr;
+  std::unique_ptr<TfLiteEngine> engine_ = nullptr;
+};
+
+// See if input tensor dims signature for height and width is -1
+// because it is so in the model.
+TEST_F(DynamicInputTest, InputHeightAndWidthMutable) {
```
lu-wang-g, line 85: I think I've commented this before. No need to test the input tensor dims for height / width of the test model. This test can be removed.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
```diff
@@ -0,0 +1,147 @@
...
+// See if output tensor has been re-dimmed as per the input
+// tensor. Expected shape: (1, input_height, input_width, 16).
+TEST_F(DynamicInputTest, OutputDimensionCheck) {
+  SUPPORT_ASSERT_OK_AND_ASSIGN(ImageData image, LoadImage("burger.jpg"));
+  std::unique_ptr<FrameBuffer> image_frame_buffer = CreateFromRgbRawBuffer(
+      image.pixel_data, FrameBuffer::Dimension{image.width, image.height});
+
+  preprocessor_->Preprocess(*image_frame_buffer);
+  absl::Status status = engine_->interpreter_wrapper()->InvokeWithoutFallback();
+  EXPECT_TRUE(status.ok());
```
lu-wang-g, line 101: This can be merged into one line, such as:
SUPPORT_ASSERT_OK(engine_->interpreter_wrapper()->InvokeWithoutFallback());

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  preprocessor_->Preprocess(*image_frame_buffer);
+
+  // Check the processed input image.
+  float *processed_input_data =
+      tflite::task::core::AssertAndReturnTypedTensor<float>(
+          engine_->GetInputs()[0]);
+
+  bool is_equal = true;
+  float epsilon = 0.1f;
+  std::string file_path =
```
lu-wang-g, line 126: How about compare `processed_input_data` with `image`? Theoretically, `image` is not changed in Preprocess(), i.e. no cropping or resizing, except that the values converted from uint8 to float.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
jonpsy: Yep, it has metadata. Since the model is `kTfLiteFloat32` type, metadata is mandatory. I used this [script](https://gist.github.com/jonpsy/500c1448fbf7434feabb5b9b804ee331)  to populate it. I've also put the issue number in the metadata.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
jonpsy: I suppose you mean `image` is the `golden image`? The problem here is, `Preprocess()` normalizes the image, which is why `processed_input_data` is a `float*`. Now, to compare, we need the "normalized" image each value would be float in range (-1, 1). 

`float` can't be stored in `jpg` so I'm storing them in `bin` file.

jonpsy: Used [this](https://gist.github.com/jonpsy/0b6b9069a3b6cb0d024c7163b30abaa6) script to normalize and store `burger.jpg`

jonpsy:
> Unable to get absolute path for `std::ifstream`. confused

Fixed, needed at add "*.bin" in`BUILD` folder of `testdata`.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
lu-wang-g: I see. You can set 
```
input_normalization.options.mean = [0]
input_normalization.options.std = [1]
```
to by-pass normalization. Then comparing with the golden image will be much easier.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
lu-wang-g: Please see the other comment about bypassing the normalization. Testing against the original image is more robust and easy to debug than comparing with the bin file.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
jonpsy: I have a question, why was this not checked before? I mean the user can do the following

```c++
// Create engine and preprocessor.
auto engine_ = absl::make_unique<TfLiteEngine>();
engine_->BuildandInit(); // Builds + init.
auto preprocessor = ImageProcessor::Create(..);

// Re-size tensor
engine_->ResizeInputTensor(1, 420, 380, 5); // Batch x height x width x channel. Notice batch and channel are invalid

// Undefined behaviour??
engine_->Preprocess(*buffer);
```

I checked `Preprocess` it claims the following:

```
    // Preprocess input image to fit model requirements.
    // For now RGB is the only color space supported, which is ensured by
    // `InitInternal`.
```

but we know that this will fail if we change after Init() right? Just as you mentioned.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
```diff
@@ -0,0 +1,131 @@
...
+    is_equal &=
+        std::fabs(static_cast<float>(*image_data) - *processed_input_data) <=
+        std::numeric_limits<float>::epsilon();
+
+  EXPECT_TRUE(is_equal);
+}
+
+// Modifying batch/depth to invalid size after Init()
+// call should throw error.
+TEST_F(DynamicInputTest, InvalidBatchOrDepthResize) {
```
lu-wang-g, line 116: Again, this won't happen, since ResizeInputTensor() should not be exposed. And we'll most likely only expose const Interpreter from engine.

lu-wang-g:
@wangtz Can you please take a look at this PR as well? Thanks!

wangtz:
Looks good. Thanks. It would be great if you could mention this behavior in  image_preprocessor.h by updating the comments.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
jonpsy: So, only `ResizeInputTensor`, but `ResizeInputTensorStrict` would be exposed via `engine_`, correct?

jonpsy: Also the comment "For now RGB is only color space support, which is ensured by 'InitInternal'" is redundant.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
jonpsy: Okay, but bear in mind that `ResizeInputTensorStrict` won't work (which we did in `height_mutable_ || width_mutable_` ) if `const` Interpreter is taken. Perhaps we could do something like this

```c++
// image_preprocessor.cc
engine_->ResizeInputTensorStrict(...)
```

where
```c++
// engine.cc
class TFLiteEngine {
public:
   ResizeInputTensorStrict(..) {
       interpreter_->ResizeInputTensorStrict(...);
}
}
```
That way we can conveniently hide `ResizeInputTensor` and expose `ResizeInputTensorStrict` all the while having `const` Interpreter.

jonpsy: Okay, I've changed it to `ResizeInputTensorStrict()` which would be exposed. Tried to resize to invaild batch size, and expected an error.

tensorflow_lite_support/cc/task/processor/image_preprocessor.cc
lu-wang-g: `ResizeInputTensorStrict` will not be exposed through `engine_wrapper_` either. But processors will have access to engine, you can still do  `ResizeInputTensorStrict` within processors.

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
lu-wang-g: I know it's a little bit confusing, so the point is end users of processors (like your tests here) will only be able to access `engine_wrapper`, not `engine`. But processors can access `engine`, because `engine` is completely internal to Task library. Therefore. You should pretend using `engine_wrapper` for these tests. And this  particular test can be removed, since it won't happen (e.g., you won't be able to  access `ResizeInputTensorStrict` through `engine_wrapper`).

lu-wang-g (commented):
Just one minor issue. We are good to check in this. Thanks for your efforts and time!!

tensorflow_lite_support/cc/test/task/processor/image_preprocessor_test.cc
jonpsy: Well, that puts things into perspective.

jonpsy:
I suppose this one is GTM?

lu-wang-g (approved):

PR #658, opened 2021-08-18
Image classifier C Wrappers

priankakariatyml:
This pull request contains changes for Image Classifier C  wrappers and basic tests. The wrappers reside in tensorflow_lite_support/c. 
Additionally, image_utils_c.h and image_utils_c.cc have been added to tensorflow_lite_support/examples/task/vision/desktop/utils. 
The BUILD file at this location is also modified accordingly.


tensorflow_lite_support/c/task/processor/BUILD
```diff
@@ -0,0 +1,14 @@
+load("@org_tensorflow//tensorflow/lite/core/shims:cc_library_with_tflite.bzl", "cc_library_with_tflite")
+
+package(
+    default_visibility = ["//tensorflow_lite_support:users"],
+    licenses = ["notice"],  # Apache 2.0
+)
+
+cc_library(
+    name = "classification_result_c_api",
```
khanhlvg, line 9: As we now have a dedicated `c` folder, let's remove the suffix `c_api` from the target and file names.

tensorflow_lite_support/c/task/processor/classification_result_c_api.cc
```diff
@@ -0,0 +1,42 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
lu-wang-g, line 1: It should be "2021". Same for the other files.

tensorflow_lite_support/c/task/processor/classification_result_c_api.h
```diff
@@ -0,0 +1,83 @@
...
+  int size;
+
+  // Array of image classifier results per image classifier head. This array can
+  // have any number of results. size holds the size of this array. size should
+  // be used to traverse this array.
+  struct Classifications *classifications;
+};
+
+// Frees up the ClassificationResult Structure.
+extern void ImageClassifierClassificationResultDelete(
```
lu-wang-g, line 76: Is extern needed here?

tensorflow_lite_support/c/task/processor/BUILD
```diff
@@ -0,0 +1,14 @@
+load("@org_tensorflow//tensorflow/lite/core/shims:cc_library_with_tflite.bzl", "cc_library_with_tflite")
```
lu-wang-g, line 1: Remove this, since it's not used?

tensorflow_lite_support/c/task/vision/BUILD
```diff
@@ -0,0 +1,42 @@
...
+
+package(
+    default_visibility = [
+        "//visibility:public",
+    ],
+    licenses = ["notice"],  # Apache 2.0
+)
+
+cc_library(
+    name = "classification_result_c_api",
```
lu-wang-g, line 11: Remove this target?

tensorflow_lite_support/c/task/processor/BUILD
priankakariatyml: Changed file names and associated build scripts. Have not removed the suffix in the names in the rule definitions for libraries in Bazel to avoid any confusion.

tensorflow_lite_support/c/task/processor/classification_result_c_api.cc
priankakariatyml: Done

tensorflow_lite_support/c/task/vision/BUILD
priankakariatyml: Done

tensorflow_lite_support/c/task/processor/BUILD
priankakariatyml: Updated

tensorflow_lite_support/c/task/vision/core/frame_buffer.h
```diff
@@ -0,0 +1,89 @@
...
+// manage the backing buffer lifecycle for the lifetime of the FrameBuffer.
+struct FrameBuffer {
+  // Colorspace format of the frame buffer.
+  enum Format format;
+  // Orientation of the frame buffer.
+  enum Orientation orientation;
+  // Dimension information for the whole frame.
+  struct Dimension dimension;
+  // Holds the stride and backing buffer for the frame buffer.
+  struct Plane plane;
```
lu-wang-g, line 82: Plane can be replaced with  `uint8_t* buffer` here.  Plane is used in the implementation details and same advanced API. We don't need to expose it here.

tensorflow_lite_support/c/task/vision/image_classifier.h
```diff
@@ -0,0 +1,189 @@
...
+extern ImageClassifier* ImageClassifierFromOptions(
+    const ImageClassifierOptions* options);
+
+// Creates ImaegeClassifier from model path and default options, returns nullptr
+// if the file doesn't exist or is not a well formatted TFLite model path.
+// This image classifier created using this function defaults to s score 
+// threshold of 0, max_results of 5. In order to override these, use:
+// ImageClassifier* ImageClassifierFromOptions(
+//   const ImageClassifierOptions* options) instead.
+extern ImageClassifier* ImageClassifierFromFile(const char* model_path);
```
lu-wang-g, line 166: ImageClassifierFromFile is not needed once you integrate BaseOptions in ImageClassifierOptions.

tensorflow_lite_support/c/task/vision/image_classifier.h
```diff
@@ -0,0 +1,189 @@
...
+// Invokes the encapsulated TFLite model and classifies the frame_buffer.
+extern struct ClassificationResult* ImageClassifierClassify(
+    const ImageClassifier* classifier, const struct FrameBuffer* frame_buffer);
+
+// Invokes the encapsulated TFLite model and classifies the
+// region of the frame_buffer specified by the bounding box.
+// Same as above, except that the classification is performed based on the
+// input region of interest. Cropping according to this region of interest is
+// prepended to the pre-processing operations.
+extern struct ClassificationResult* ImageClassifierClassifyWithBoundingBox(
```
lu-wang-g, line 177: ImageClassifierClassifyWithBoundingBox => ImageClassifierClassifyWithRoi. ROI: region of interest

tensorflow_lite_support/c/task/vision/core/frame_buffer.h
priankakariatyml: fixed

tensorflow_lite_support/c/task/processor/classification_result.h
```diff
@@ -0,0 +1,83 @@
...
+  int size;
+
+  // Array of image classifier results per image classifier head. This array can
+  // have any number of results. size holds the size of this array. size should
+  // be used to traverse this array.
+  struct Classifications *classifications;
+};
+
+// Frees up the ClassificationResult Structure.
+extern void ImageClassifierClassificationResultDelete(
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/processor/classification_result.h
```diff
@@ -0,0 +1,83 @@
...
+  // the TFLite Model Metadata [1].
+  //
+  // [1]: https://www.tensorflow.org/lite/convert/metadata
+  int index;
+
+  // The score for this class e.g. (but not necessarily) a probability in [0,1].
+  float score;
+
+  // A human readable name of the class filled from the label map.
+  char *display_name;
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/processor/classification_result.cc
```diff
@@ -0,0 +1,49 @@
...
+
+#include "tensorflow_lite_support/c/task/processor/classification_result.h"
+
+#include <memory>
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+void ImageClassifierClassificationResultDelete(struct ClassificationResult* classification_result) {
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/processor/classification_result.h
```diff
@@ -0,0 +1,83 @@
...
+  // An ID for the class, not necessarily human-readable (e.g. a Google
+  // Knowledge Graph ID [1]), filled from the label map.
+  //
+  // [1]: https://developers.google.com/knowledge-graph
+  char *class_name;
+  
+};
+
+// List of predicted classes (aka labels) for a given image classifier head.
+struct Classifications {
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/processor/BUILD
```diff
@@ -0,0 +1,12 @@
+package(
+    default_visibility = ["//tensorflow_lite_support:users"],
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/vision_common.h
```diff
@@ -0,0 +1,47 @@
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/vision_common.h
```diff
@@ -0,0 +1,47 @@
...
+//
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+// Holds the region of interest used for image classification.
+struct BoundingBox {
+  // The X coordinate of the top-left corner, in pixels.
+  int32_t origin_x;
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/core/frame_buffer.h
```diff
@@ -0,0 +1,89 @@
...
+// --------------------------------------------------------------------------
+/// C Struct for Holding the FrameBuffer.
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+// Colorspace formats.
+enum Format { kRGBA, kRGB, kNV12, kNV21, kYV12, kYV21, kGRAY, kUNKNOWN};
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,217 @@
...
+
+#include "absl/strings/string_view.h"
+#include "absl/strings/str_format.h"
+#include "tensorflow_lite_support/cc/task/vision/image_classifier.h"
+#include "tensorflow_lite_support/cc/task/vision/proto/image_classifier_options_proto_inc.h"
+#include "tensorflow_lite_support/cc/task/vision/proto/classifications_proto_inc.h"
+#include "tensorflow_lite_support/c/task/vision/utils/frame_buffer_cpp_c_utils.h"
+
+using ::tflite::support::StatusOr;
+using ClassificationResultCPP = ::tflite::task::vision::ClassificationResult;
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.h
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,217 @@
...
+#include <memory>
+
+#include "absl/strings/string_view.h"
+#include "absl/strings/str_format.h"
+#include "tensorflow_lite_support/cc/task/vision/image_classifier.h"
+#include "tensorflow_lite_support/cc/task/vision/proto/image_classifier_options_proto_inc.h"
+#include "tensorflow_lite_support/cc/task/vision/proto/classifications_proto_inc.h"
+#include "tensorflow_lite_support/c/task/vision/utils/frame_buffer_cpp_c_utils.h"
+
+using ::tflite::support::StatusOr;
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/processor/classification_result.h
```diff
@@ -0,0 +1,83 @@
...
+
+  // The array of predicted classes, usually sorted by descending scores (e.g.
+  // from high to low probability). Since this array is dynamically allocated, 
+  // use size to traverse through the array.
+  struct Class *classes; 
+};
+  
+// Holds Image Classification results. 
+// Contains one set of results per image classifier head.
+struct ClassificationResult {
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/processor/BUILD
```diff
@@ -0,0 +1,12 @@
...
+    licenses = ["notice"],  # Apache 2.0
+)
+
+cc_library(
+    name = "classification_result_c_api",
+    srcs = [
+        "classification_result.cc",
+    ],
+    hdrs = ["classification_result.h"],
+)
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.h
```diff
@@ -0,0 +1,189 @@
...
+/// struct ClassificationResult *classification_result = 
+///   ImageClassifierClassify(classifier, &frame_buffer,);
+///
+/// // Dispose of the API object.
+/// ImageClassifierDelete(classifier);
+/// 
+/// Advanced Usage with Options:
+/// <pre><code>
+/// // Create the interpreter options
+/// ImageClassifierOptions *options = ImageClassifierOptionsCreate();
```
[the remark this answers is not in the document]
priankakariatyml: @khanhlvg I initially used a simple struct but decided against it later. I had shared a doc when I implemented this with my reasoning. Image Classifier options have arrays like class name whitelist and black list which are of unknown sizes. We can't set the pointer to the base address of these arrays in the respective C++ options. We have to loop through them and add them into the respective C++ options in the proto buf. For that we need the length of these arrays. If we are going with the struct approach we will have to add members for the sizes of these options. Another way was to let users of the C apis add options via methods. Got the idea from [TfLiteInterpreterOptions](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/c/c_api.h)
I felt the latter approach is better since it eliminates addition of unwanted size members in the struct. 
A different strategy would be to have a nested struct type like say struct StringArrayOption within Classifier Options with size and pointer to array as members and let all string array options be of that type. If that's okay we can change this. Or if you have another workaround or approach, please let me know.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,223 @@
...
+  StatusOr<ClassificationResultCpp> classification_result_cpp =
+      classifier->impl->Classify(*cpp_frame_buffer, cc_roi);
+
+  if (!classification_result_cpp.ok()) return nullptr;
+
+  TfLiteClassificationResult* c_classification_result =
+      GetClassificationResultCStruct(
+          ClassificationResultCpp(classification_result_cpp.value()));
+
+  return c_classification_result;
```
lu-wang-g, line 214: Line 210 - 214  can be merged to 
```
return GetClassificationResultCStruct(classification_result_cpp.value());
```

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,223 @@
...
+}
+
+TfLiteClassificationResult* TfLiteImageClassifierClassifyWithRoi(
+    const TfLiteImageClassifier* classifier,
+    const TfLiteFrameBuffer* frame_buffer, const TfLiteBoundingBox* roi) {
```
lu-wang-g, line 196: These function input params cannot be null. So either check if they're null, or change pointer to reference.

Same for all other functions.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,223 @@
...
+  BoundingBoxCpp cc_roi;
+  cc_roi.set_origin_x(roi->origin_x);
+  cc_roi.set_origin_y(roi->origin_y);
+  cc_roi.set_width(roi->width);
+  cc_roi.set_height(roi->height);
+
+  std::unique_ptr<FrameBufferCpp> cpp_frame_buffer =
+      CreateCppFrameBuffer(frame_buffer);
+  StatusOr<ClassificationResultCpp> classification_result_cpp =
+      classifier->impl->Classify(*cpp_frame_buffer, cc_roi);
```
lu-wang-g, line 206: What if `cpp_frame_buffer` is null?

tensorflow_lite_support/c/task/vision/utils/frame_buffer_cpp_c_utils.cc
```diff
@@ -0,0 +1,40 @@
...
+namespace {
+using FrameBufferCpp = ::tflite::task::vision::FrameBuffer;
+}
+
+std::unique_ptr<FrameBufferCpp> CreateCppFrameBuffer(
+    const TfLiteFrameBuffer* frame_buffer) {
+  FrameBufferCpp::Format frame_buffer_format =
+      FrameBufferCpp::Format((*frame_buffer).format);
+
+  auto cpp_frame_buffer = tflite::task::vision::CreateFromRawBuffer(
```
lu-wang-g, line 29: You can just return `StatusOr<std::unique_ptr<FrameBufferCpp>>`.  Then in Classify(), check the status of the return to verify if FrameBufferCpp is created successfully. 

Here,  the code becomes much simpler by:
```
return tflite::task::vision::CreateFromRawBuffer(
     frame_buffer->buffer,
      {frame_buffer->dimension.width, frame_buffer->dimension.height},
      frame_buffer_format);
```
```

File: tensorflow_lite_support/c/task/vision/utils/frame_buffer_cpp_c_utils.h [hunk 26]
```diff
@@ -0,0 +1,29 @@
...
+
+#include "tensorflow_lite_support/c/task/vision/core/frame_buffer.h"
+#include "tensorflow_lite_support/cc/task/vision/utils/frame_buffer_common_utils.h"
+
+// Utils for Conversions between C and C++ FrameBuffer
+// -----------------------------------------------------------------
+// Meant to be used with vision C apis.
+
+// Creates the C++ FrameBuffer from the C FrameBuffer
+std::unique_ptr<::tflite::task::vision::FrameBuffer> CreateCppFrameBuffer(
```
lu-wang-g, line 26: Add namespace for C++ headers.

tensorflow_lite_support/c/task/vision/image_classifier.h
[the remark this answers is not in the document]
khanhlvg: As Lu's suggestion in her comment in the doc, let's create a TfLiteStringOption struct that contains an array and an int to store the array size.

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: AFAIK, C does not have pass by reference. I will change the private methods used by the C++ code. Not the ones that are declared in C header files. Also, I thought null check for function params would be overkill. Since there is a chance that the programmer can pass an invalid address and still get a seg fault. A null check won't prevent that. Followed the pattern of params in [nl_classifier](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/c/task/text/nl_classifier.cc#L69-L72)

I'll still add the null check for params when I push the fixes for these comments and we can keep it that way for vision. But since it won't prevent a seg fault from happening because of invalid addresses, we may consider adding a note about that in the documentation.
C APIs are meant to be used with wrappers. So the onus can be on the caller framework to not pass an explicit null or an invalid address.

priankakariatyml:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here with `@googlebot I signed it!` and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Ftensorflow%2Ftflite-support%2Fpull%2F658) for more info**.

@googlebot I signed it!

tensorflow_lite_support/c/task/vision/utils/frame_buffer_cpp_c_utils.cc
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: Fixed. Now CreateCppFrameBuffer returns StatusOr

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,223 @@
...
+
+  if (!classification_result_cpp.ok()) return nullptr;
+
+  TfLiteClassificationResult* c_classification_result =
+      GetClassificationResultCStruct(
+          ClassificationResultCpp(classification_result_cpp.value()));
```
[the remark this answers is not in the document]
priankakariatyml: Fixed. Changed `ClassificationResultCpp` to pass by reference `GetClassificationResultCStruct()` and passing `classification_result_cpp.value()`

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,223 @@
...
+
+  StatusOr<ClassificationResultCpp> classification_result_cpp =
+      classifier->impl->Classify(*cpp_frame_buffer);
+
+  if (!classification_result_cpp.ok()) return nullptr;
+
+  TfLiteClassificationResult* c_classification_result =
+      GetClassificationResultCStruct(
+          ClassificationResultCpp(classification_result_cpp.value()));
+
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,223 @@
...
+TfLiteImageClassifier* TfLiteImageClassifierFromFile(const char* model_path) {
+  TfLiteImageClassifierOptions* default_options =
+      TfLiteImageClassifierOptionsDefault();
+  TfLiteImageClassifierOptionsSetModelFilePath(default_options, model_path);
+  return TfLiteImageClassifierFromOptions(default_options);
+}
+
+TfLiteClassificationResult* GetClassificationResultCStruct(
+    const ClassificationResultCpp classification_result_cpp) {
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,223 @@
...
+TfLiteImageClassifier* TfLiteImageClassifierFromOptions(
+    const TfLiteImageClassifierOptions* options) {
+  auto classifier_status =
+      ImageClassifierCpp::CreateFromOptions(*(options->impl));
+
+  if (classifier_status.ok()) {
+    return new TfLiteImageClassifier{
+        .impl = std::unique_ptr<ImageClassifierCpp>(
+            dynamic_cast<ImageClassifierCpp*>(
+                classifier_status.value().release()))};
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/core/frame_buffer.h
```diff
@@ -0,0 +1,81 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+#ifndef TENSORFLOW_LITE_SUPPORT_C_TASK_VISION_FRAME_BUFFER_H_
+#define TENSORFLOW_LITE_SUPPORT_C_TASK_VISION_FRAME_BUFFER_H_
+
+#include <stdint.h>
+
+// Defines C structs for Holding the frame buffer.
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/BUILD
```diff
@@ -0,0 +1,28 @@
...
+
+cc_library_with_tflite(
+    name = "image_classifier",
+    srcs = [
+        "image_classifier.cc",
+    ],
+    hdrs = [
+        "image_classifier.h",
+        # "//tensorflow_lite_support/c/task/processor:classification_result",
+        # "//tensorflow_lite_support/c/task/processor:bounding_box"
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/processor/classification_result.cc
```diff
@@ -0,0 +1,47 @@
...
+    TfLiteClassifications classifications =
+        classification_result->classifications[head];
+    for (int rank = 0; rank < classifications.size; ++rank) {
+      // `strdup` obtains memory using `malloc` and the memory needs to be
+      // released using `free`.
+      free(classifications.categories[rank].display_name);
+      free(classifications.categories[rank].label);
+    }
+
+    if (classifications.size >= 1) delete[] classifications.categories;
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/processor/classification_result.cc
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  if (classification_result->size >= 1)
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.h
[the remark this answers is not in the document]
priankakariatyml: Implemented

tensorflow_lite_support/c/task/vision/image_classifier.h
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: Added null check for params. Changed all methods to pass by reference sparing the public C APIs.

lu-wang-g: Yeah, you're right. Reference is not supported  in C.

The main concern for null pointer is that, Task library can return null ImageClassifier and FrameBuffer, therefore, users may just pass whatever they get to other methods of Task Library. In the future, we should also implementing  error reporting, so that the error message can be properly surfaced.

tensorflow_lite_support/c/task/processor/base_options.h
```diff
@@ -0,0 +1,77 @@
...
+#include <stdint.h>
+
+// Defines C Structs for Base Options Shared by all tasks.
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+// Holds settings to configure cpu execution.
+typedef struct TfLiteCpuSettings {
```
lu-wang-g, line 27: Thanks for implementing these structs! 

Just a reminder, please don't add more features in this PR. It is already very large. In general, small PRs (less than 5 files and less than 300 lines of code) are preferred. It helps reduce complexity in both development and code review. In the future, we can try adding code in the following order.
* APIs: methods declarations (.h files), protos, BUILD targets.
* Implementations: .cc files and some simple smoke tests.
* More thorough tests.

tensorflow_lite_support/c/task/processor/base_options.h
```diff
@@ -0,0 +1,77 @@
...
+// Holds settings to configure TFLite runtime.
+typedef struct TfLiteTflSettings {
+  // Specifies how to configure cpu execution.
+  TfLiteCpuSettings cpu_settings;
+} TfLiteTflSettings;
+
+// Holds settings for one possible acceleration configuration.
+typedef struct TfLiteComputeSettings {
+  // Holds settings to configure TFLite runtime.
+  TfLiteTflSettings tflite_settings;
```
lu-wang-g, line 44: TFLiteSettings can be removed. Other fields in ComputeSettings is not used, therefore. Please just expose the everything under TFLiteSettings under ComputeSettings. See the Java API for references. https://github.com/tensorflow/tflite-support/blob/1ad1e531d7cf439c200f9df3ae87009d986a6f52/tensorflow_lite_support/java/src/java/org/tensorflow/lite/task/core/BaseOptions.java

lu-wang-g: Note that, ExternalFile is not added in the Java API yet. But I think what you did here (TfLiteExternalFile) looks good.

tensorflow_lite_support/c/task/processor/base_options.h
```diff
@@ -0,0 +1,77 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+#ifndef TENSORFLOW_LITE_SUPPORT_C_TASK_PROCESSOR_BASE_OPTIONS_H_
+#define TENSORFLOW_LITE_SUPPORT_C_TASK_PROCESSOR_BASE_OPTIONS_H_
+
+#include <stdint.h>
+
+// Defines C Structs for Base Options Shared by all tasks.
```
lu-wang-g, line 20: Please put this file under c/task/core.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,225 @@
...
+        ->set_num_threads(-1);
+  }
+
+  if (c_options->classification_options.class_name_blacklist.length > 0 &&
+      c_options->classification_options.class_name_whitelist.length > 0)
+    return nullptr;
+
+  if (c_options->classification_options.class_name_blacklist.length > 0 &&
+      c_options->classification_options.class_name_whitelist.length)
+    return nullptr;
```
lu-wang-g, line 82: These checks can be removed, because they will be checked and verified in the C++ layer.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,225 @@
...
+  
+  if (c_options->classification_options.class_name_blacklist.length > 0) {
```
lu-wang-g, line 84: This if check is redundant. If length <0, the for loop below won't execute.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,225 @@
...
+    for (int i = 0;
+         i < c_options->classification_options.class_name_blacklist.length; i++)
+      cpp_options->add_class_name_blacklist(
+          c_options->classification_options.class_name_blacklist.list[i]);
+  }
+  
+  if (c_options->classification_options.class_name_whitelist.length >
```
lu-wang-g, line 91: Same here. Remove the if check.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,225 @@
...
+      cpp_options->add_class_name_whitelist(
+          c_options->classification_options.class_name_whitelist.list[i]);
+  }
+
+  if (c_options->classification_options.display_names_local) {
+    cpp_options->set_display_names_locale(
+        c_options->classification_options.display_names_local);
+  }
+
+  if (c_options->classification_options.max_results > 0) {
```
lu-wang-g, line 104: If check can be removed. Just rely on the  C++ layer to check.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -0,0 +1,225 @@
...
+    cpp_options->set_max_results(c_options->classification_options.max_results);
+  }
+
+  if (c_options->classification_options.score_threshold >= 0) {
```
lu-wang-g, line 108: Same here. Remove the if check.

tensorflow_lite_support/c/task/vision/image_classifier.h
```diff
@@ -0,0 +1,103 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+typedef struct TfLiteImageClassifier TfLiteImageClassifier;
+
+typedef struct TfLiteImageClassifierOptions {
+  TfLiteClassificationOptions classification_options;
+  TfLiteBaseOptions base_options;
+
```
lu-wang-g, line 63: nit: please remove the empty line of line 63.

lu-wang-g:
@priankakariatyml as discussed with Khanh today, we need to implement the [ErrorReportor](https://github.com/tensorflow/tensorflow/blob/c49813dd0f4af5e74bad512ff58cbc22c81a4d39/tensorflow/lite/c/c_api.h#L141) similar to TFLite C API, so that error messages can be properly surfaced from the C++ layer. Let's work on the ErrorReporter after this PR is checked in.

Also, thanks a lot for your patience in keeping iterating the PR. I hope the process doesn't frustrated you that much. In Google, we'll go through a much more restrict code review process. It's quite common to iterate a PR of like 200 lines many times. And since you're new to Google's code stack, it's very impressive that you can quickly ramp up and being productive. Really appreciate your corporation and contributions!

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: We can't remove this check. The C `TfLiteImageClassifierOptions` needs to be zero initialised by users of C api to prevent garbage values in members. Otherwise it will result in seg faults in C and C++.  I have mentioned this in the documentation. `max_results` gets initialised to 0 by default in this case. Setting 0 for `max_results` in cpp layer results in a not ok status for creation of classifier. This may not be what the user of the C API intended to do. 

Ideally from C, you should be able to set only options you need and the underlying layers should take care of initialising others to default values. IMO the CPP layer should use default `max_results` when it is 0. Until then we will need this check given we can't have any defaults for struct members in C. 

Another workaround is to have a function in the C layer to return a default `TfLiteImageClassifierOptions` set with defaults and mandate that C options should be created only using this function. IMO this needs to be done. Then the programmer can set the options to whatever values he wants and also not worry about zero initialisation. In this case we can avoid these checks completely since the onus is on them to correctly change values.

Same applies for `num_threads`. We can remove the check for `score_threshold` since 0 is the expected default. Garbage values shouldn't be the frameworks headache.

tensorflow_lite_support/c/task/processor/base_options.h
priankakariatyml: Moved both base_options.h and classification_options.h to c/task/core

tensorflow_lite_support/c/task/processor/base_options.h
priankakariatyml: I have removed all the nested structs in `TfLiteComputeSettings`. In Java, the `num_threads` is in Base Options. But I have maintained `num_threads` as a member of `TfLiteComputeSettings` since that feels more appropriate given the structure of the lite options. Later acceleration related options can also be added to `TfLiteComputeSettings`.

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: Removed for blacklist and whitelist. I tested the CPP implementation with various values of `num_threads`. Please see my reason for not removing the check for ` num_threads` [here]( https://github.com/tensorflow/tflite-support/pull/658#discussion_r704151228). It's a similar scenario.

priankakariatyml:
> @priankakariatyml as discussed with Khanh today, we need to implement the [ErrorReportor](https://github.com/tensorflow/tensorflow/blob/c49813dd0f4af5e74bad512ff58cbc22c81a4d39/tensorflow/lite/c/c_api.h#L141) similar to TFLite C API, so that error messages can be properly surfaced from the C++ layer. Let's work on the ErrorReporter after this PR is checked in.
> 
> Also, thanks a lot for your patience in keeping iterating the PR. I hope the process doesn't frustrated you that much. In Google, we'll go through a much more restrict code review process. It's quite common to iterate a PR of like 200 lines many times. And since you're new to Google's code stack, it's very impressive that you can quickly ramp up and being productive. Really appreciate your corporation and contributions!

Thanks @lu-wang-g . I don't have a problem. We can iterate until it's ready. I have fixed your comments for today and left some replies. Do check those. I need your opinion on one aspect of the struct options. I will also elaborate

tensorflow_lite_support/c/task/vision/image_classifier.cc
lu-wang-g: I agree we should implementing a function to return the default settings. Please use the same [default value](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/cc/task/processor/proto/classification_options.proto) as defined in the C++ layer.

lu-wang-g: And also a function to delete/free the setting struct?

tensorflow_lite_support/c/task/vision/image_classifier.cc
lu-wang-g: score_threshold is a bit different. If you set 0, then everything below 0 will be filtered. See the code [here](  https://github.com/tensorflow/tflite-support/blob/17cdc506a620e3a0745a11023ebc2542a5237d25/tensorflow_lite_support/cc/task/processor/classification_postprocessor.cc#L205). How about setting the default to FLT_MAX? See https://en.cppreference.com/w/c/types/limits.

tensorflow_lite_support/c/task/core/base_options.h
```diff
@@ -0,0 +1,67 @@
...
+extern "C" {
+#endif  // __cplusplus
+
+// Holds settings for one possible acceleration configuration.
+typedef struct TfLiteComputeSettings {
+  // Specifies the number of threads to be used for TFLite
+  // ops that support multi-threading when running inference with CPU.
+  // num_threads should be greater than 0 or equal to -1. Setting num_threads to
+  // -1 has the effect to let TFLite runtime set the value.
+  int num_threads;
```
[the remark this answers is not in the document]
khanhlvg: If only parts of the model can run on a delegate, wouldn't num_threads still control how many threads used to run the ops that aren't delegated? If that's the case, I think we shouldn't make CPUSettings and GPUSettings mutually exclusive. What about putting num_threads directly under TfLiteComputeSettings?

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: Score threshold default should be 0 right? FLT_MAX would filter everything out as score is probabilities. By default everything should be returned IMO.
max_results default can be  -1. [Doc](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/cc/task/processor/proto/classification_options.proto#L27-L30) says so.

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: Yeah if framework is creating the default struct, a free is needed. Do you want me to update this in this PR. Or once this is merged as most tests would also change. Won't be a very big change though.

tensorflow_lite_support/c/task/core/base_options.h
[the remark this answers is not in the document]
lu-wang-g: Good point, Khanh! I double checked the code, num_threads is applied no matter if other delegates are chosen.  And CPUSettings / other delegate settings are not mutually exclusive; other delegates are mutually exclusive. See the [delegate options](https://github.com/tensorflow/tensorflow/blob/2a1b1da5f3625a5d5b996b6b508768b85331a836/tensorflow/lite/experimental/acceleration/configuration/configuration.proto#L52), CPU is not included. So let's just follow the C++ structure for CPUSettings. What do you think?

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
```diff
@@ -0,0 +1,302 @@
...
+  TfLiteClassificationResult* classification_result =
+      TfLiteImageClassifierClassify(image_classifier, &frame_buffer);
+
+  CImageDataFree(&image_data);
+
+  ASSERT_NE(classification_result, nullptr) << "Classification Result is NULL";
+  EXPECT_TRUE(classification_result->size >= 1)
+      << "Classification Result size is 0";
+  EXPECT_NE(classification_result->classifications, nullptr)
+      << "Classification Result Classifications is NULL";
```
lu-wang-g, line 139: Those messages look redundant. The `EXPECT_XXX` statements are already self-explained. You can remove these messages in this file.

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
```diff
@@ -0,0 +1,302 @@
...
+CImageData LoadImage(const char* image_name) {
+  return CDecodeImageFromFile(
+      JoinPath("./" /*test src dir*/, kTestDataDirectory, image_name).data());
+}
+
+TEST(CImageClassifierFromOptionsTest, FailsWithMissingModelPath) {
+  TfLiteImageClassifierOptions options = {0};
+  TfLiteImageClassifier* image_classifier =
+      TfLiteImageClassifierFromOptions(&options);
+  ASSERT_EQ(image_classifier, nullptr);
```
lu-wang-g, line 49: Use EXPECT. ASSERT_EQ will make the test fail. 

Have you tried running this test? i.e. `bazel test tensorflow_lite_support/c/test/task/vision:image_classifier_test`?

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
```diff
@@ -0,0 +1,302 @@
...
+  TfLiteImageClassifierOptions options = {0};
+  options.base_options.model_file.file_path = model_path;
+
+  char* class_name_blacklist[] = {"brambling"};
+  options.classification_options.class_name_blacklist.list =
+      class_name_blacklist;
+  options.classification_options.class_name_blacklist.length = 1;
+
+  char* class_name_whitelist[] = {"cheeseburger"};
+  options.classification_options.class_name_whitelist.list =
```
lu-wang-g, line 95: Forgot to mention, this `white/blacklist` need to be renamed to `allow/denylist`  in our new API. It's OK to fix it in another PR. Please add a todo in image_classifier_options.h, such as:
```
TODO(prianka): change `white/blacklist` to `allow/denylist`.
```

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
```diff
@@ -0,0 +1,302 @@
...
+  CImageDataFree(&image_data);
+
+  ASSERT_NE(classification_result, nullptr) << "Classification Result is NULL";
+  EXPECT_TRUE(classification_result->size >= 1)
+      << "Classification Result size is 0";
+  EXPECT_NE(classification_result->classifications, nullptr)
+      << "Classification Result Classifications is NULL";
+  EXPECT_TRUE(classification_result->classifications->size >= 1)
+      << "Classification Result Classifications Size is 0";
+  EXPECT_NE(classification_result->classifications->categories, nullptr)
```
lu-wang-g, line 171: Need to check the score and labels. Same for the other "Succeeds" tests. Again, OK to add the tests in another PR. Please add a TODO here.

lu-wang-g (commented):
The code looks great now! Just had a few more very minor comments. Please add those TODOs for reminder and then we can check in this PR. The default classification_options can also be implemented in a following PR.

tensorflow_lite_support/c/task/core/base_options.h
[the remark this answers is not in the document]
khanhlvg: SGTM. Let's follow the same structure as the C++.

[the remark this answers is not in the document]
priankakariatyml: Does that mean it still should be in TfliteCpuSettings or directly in TfLiteBaseOptions like in C++?

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
priankakariatyml: Yes. It works fine. If ASSERT_EQ fails, it aborts only the current test. If it succeeds, it continues with the rest of the code in the function. Since there was no cleanup code after the ASSERT, didn't make sense to me to use EXPECT_EQ. Since this is a fails case for missing model path, ideally the image_classifier would never be non null.  Hence didn't add any cleanup. But logically, we should use EXPECT_EQ and delete the image_classifier if it is non null after. Let me know if this should be done.

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
priankakariatyml: Added TODOs

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
priankakariatyml: Added TODOs

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
priankakariatyml: Removed messages

tensorflow_lite_support/c/task/core/base_options.h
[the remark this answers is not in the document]
lu-wang-g: Please put it in TfliteCpuSettings. Thanks!

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
lu-wang-g: ASSERT works in this case. But theoretically, you want to use EXPECT to validate an invariant for your testing purposes, while using ASEERT to check invariants that are critical to execute the rest of your test. In this case, image_classifier is a test invariant that you want to check, so use EXPECT makes more sense.  Once you implemented error_status, you may also want to verify the error message and error code in this test. Therefore using ASSERT is inappropriate.

lu-wang-g (commented):
We're having an issue to merge external PRs. A fix will be checked in today. Once the fix is alive, I'll merge this PR.

tensorflow_lite_support/c/task/core/base_options.h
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
priankakariatyml: Got it. I have fixed this. But haven't added the cleanup for this test in the update since the structure of test will change to verifying error codes rather than nullptr once it's added. Updated tests run fine from my end.

lu-wang-g (approved):

lu-wang-g (approved):

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: @lu-wang-g By default variable sized members like class name black list, white list, and strings like display_name_local will be initialised to null pointers when we provide method to return a default initialised struct. In this case, client can allocate them using malloc or just assign address of a variable on stack. If we return a pointer to the options, we will have to provide a free method and we wouldn't know how the pointer members like strings, string arrays were allocated and hence can't simply call a delete on these members if they are passed in to the C library for deallocation. Wouldn't it be better to just return an options struct by value when client calls TfLiteImageClassifierOptionsCreate initialized with default values and avoid providing a free method. We can leave the alloc and free of pointer members to the client like how it is currently and still provide them a default initialised options which will avoid garbage values and redundant checks in C code.

lu-wang-g: Yeah, I think that makes sense. Please create TfLiteImageClassifierOptionsCreate and let it return an options struct object. Thanks for the discussion!

PR #679, opened 2021-09-09
Add Image Transformer Library

jonpsy:
@lu-wang-g @wangtz 

**TODO**

- [x] `ImageTransformerOptions.proto` files and fill in `BUILD`.
- [x] `PostProcess` logic using a class derived from  `PostProcessor`.
- [x]  Ensure dynamic image support exist.
- [x]  `.h` and `.cc` file for `ImageTransformer`.
- [x] Use `Postprocessor` class to delegate the task.
- [ ]  Add test and confirm working.
- [ ] Write a demo app.


tensorflow_lite_support/cc/task/vision/proto/BUILD
```diff
@@ -242,3 +242,31 @@ cc_library(
...
+)
+
+cc_library(
+    name = "image_transformer_options_proto_inc",
+    hdrs = ["image_transformer_options_proto_inc.h"],
+    deps = [
+        ":image_transformer_options_cc_proto",
+        "//tensorflow_lite_support/cc/task/core/proto:external_file_proto_inc",
+    ],
+)
```
lu-wang-g, line 272: An empty line, please.

tensorflow_lite_support/cc/task/vision/proto/image_transformer_options.proto
```diff
@@ -0,0 +1,69 @@
...
+import "tensorflow/lite/experimental/acceleration/configuration/configuration.proto";
+import "tensorflow_lite_support/cc/task/core/proto/base_options.proto";
+import "tensorflow_lite_support/cc/task/core/proto/external_file.proto";
+
+// Options for setting up an ObjectDetector.
+// Next Id: 10.
+message ImageTransformerOptions {
+  // Base options for configuring Task library, such as specifying the TfLite
+  // model file with metadata, accelerator options, etc.
+  optional tflite.task.core.BaseOptions base_options = 9;
```
lu-wang-g, line 29: Field id should just be 1, 2 , 3, ....

tensorflow_lite_support/cc/task/vision/proto/image_transformer_options.proto
```diff
@@ -0,0 +1,69 @@
...
+
+syntax = "proto2";
+
+package tflite.task.vision;
+
+import "tensorflow/lite/experimental/acceleration/configuration/configuration.proto";
+import "tensorflow_lite_support/cc/task/core/proto/base_options.proto";
+import "tensorflow_lite_support/cc/task/core/proto/external_file.proto";
+
+// Options for setting up an ObjectDetector.
```
lu-wang-g, line 24: ObjectDetector => ImageTransformer

tensorflow_lite_support/cc/task/vision/proto/image_transformer_options.proto
```diff
@@ -0,0 +1,69 @@
...
+
+  // Legacy method for specifying the TFLite model file, as a single standalone
+  // TFLite file packed with TFLite Model Metadata [1]. Those are mandatory, and
+  // used to populate e.g. the label map and recommended score threshold.
+  //
+  // [1]: https://www.tensorflow.org/lite/convert/metadata
+  //
+  // Deprecated: prefer using `base_options.model_file`, which is mutually
+  // exclusive with this field.
+  optional core.ExternalFile model_file_with_metadata = 1;
```
lu-wang-g, line 39: This  doesn't need. model_file can be set using BaseOptions.

tensorflow_lite_support/cc/task/vision/proto/image_transformer_options.proto
```diff
@@ -0,0 +1,69 @@
...
+  // Legacy method for specifying the number of threads to be used for TFLite
+  // ops that support multi-threading when running inference with CPU.
+  // num_threads should be greater than 0 or equal to -1. Setting num_threads to
+  // -1 has the effect to let TFLite runtime set the value.
+  //
+  // Deprecated: only works with `model_file_with_metadata`. Prefer using
+  // `base_options` to specifying the TFLite model and using
+  // `base_options.compute_settings.tflite_settings.cpu_settings.num_threads`,
+  // to configure the number of threads.
+  optional int32 num_threads = 7 [default = -1];
```
lu-wang-g, line 50: num_threads is configured through compute_settings, so can be removed  too.

tensorflow_lite_support/cc/task/vision/image_transformer.cc
```diff
@@ -0,0 +1,195 @@
...
+  if (num_input_models != 1) {
+    return CreateStatusWithPayload(
+        StatusCode::kInvalidArgument,
+        absl::StrFormat("Expected exactly one of `base_options.model_file` or "
+                        "`model_file_with_metadata` to be provided, found %d.",
+                        num_input_models),
+        TfLiteSupportStatus::kInvalidArgumentError);
+  }
+  if (options.base_options().compute_settings().tflite_settings().cpu_settings().num_threads() == 0 ||
+      options.base_options().compute_settings().tflite_settings().cpu_settings().num_threads() < -1) {
```
jonpsy, line 99: Is this the correct way to get `num_threads`?

lu-wang-g: Yes, that's correct. But you don't need to check num_threads specifically. It has been verified [here](https://github.com/tensorflow/tflite-support/blob/765e82607c6a85ac9cdbda12467ff1f56826914f/tensorflow_lite_support/cc/task/core/task_api_factory.h#L133-L141) when the base task is created.

tensorflow_lite_support/cc/task/vision/image_transformer.cc
```diff
@@ -0,0 +1,195 @@
...
+  if (output_tensor->dims->data[0] != 1) {
+    return CreateStatusWithPayload(
+        StatusCode::kInvalidArgument,
+        absl::StrFormat("Expected batch size of 1, found %d.",
+                        output_tensor->dims->data[0]),
+        TfLiteSupportStatus::kInvalidOutputTensorDimensionsError);
+  }
+  // TODO: Will the output be float and should be converted or directly available?
+  // The example had float and it had to be converted. Anyway, we're guaranteed to have uint8 as output.
+  has_uint8_outputs_ = (output_tensor->type == kTfLiteUInt8);
```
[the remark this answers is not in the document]
lu-wang-g: I think both situation will happen. If the output is float, it most likely will be in the same range as the input image, meaning if the input is normalized to [0, 1], you need to denormalized the output from [0, 1] to [0,  255]. We'll need to implement for both. See the style transfer model here: https://www.tensorflow.org/lite/examples/style_transfer/overview#performance_benchmarks

tensorflow_lite_support/cc/task/vision/image_transformer.cc
```diff
@@ -0,0 +1,226 @@
...
+}
+
+StatusOr<FrameBuffer> ImageTransformer::Transform(
+    const FrameBuffer& frame_buffer, const BoundingBox& roi) {
+  return InferWithFallback(frame_buffer, roi);
+}
+
+StatusOr<FrameBuffer> ImageTransformer::Postprocess(
+    const std::vector<const TfLiteTensor*>& /*output_tensors*/,
+    const FrameBuffer& /*frame_buffer*/, const BoundingBox& /*roi*/) {
```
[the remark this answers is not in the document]
jonpsy: As discussed, that's acceptable.

jonpsy:
Thoughts on testing @lu-wang-g ?

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
```diff
@@ -0,0 +1,135 @@
...
+  // Whether the model features quantized inference type (QUANTIZED_UINT8). This
+  // is currently detected by checking if all output tensors data type is uint8.
+  bool has_uint8_outputs_;
+
+  std::unique_ptr<vision::NormalizationOptions> options_;
+
+  absl::Status Init(std::unique_ptr<vision::NormalizationOptions> options);
+};
+
+absl::StatusOr<vision::FrameBuffer> ImagePostprocessor::Postprocess() {
```
lu-wang-g, line 70: The implementation should be moved to the .cc file. Only template method is allowed to be implemented in the header.

lu-wang-g:
You can have an end-2-end test using ImageTransformer to validate the implementation. Try the style transfer model here: https://www.tensorflow.org/lite/examples/style_transfer/overview#performance_benchmarks

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
```diff
@@ -0,0 +1,135 @@
...
+//    - batch inference is not supported (`batch` is required to be 1).
+//    - only RGB inputs are supported (`channels` is required to be 3).
+//    - if type is kTfLiteFloat32, NormalizationOptions are required to be
+//      attached to the metadata for input normalization.
+class ImagePostprocessor : public Postprocessor {
+ public:
+  static tflite::support::StatusOr<std::unique_ptr<ImagePostprocessor>>
+  Create(core::TfLiteEngine* engine,
+         const std::initializer_list<int> output_indices,
+         std::unique_ptr<vision::NormalizationOptions> options);
```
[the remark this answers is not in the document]
jonpsy: Okay, but won't this break the existing design pattern? To implement this, I think we can extract input metadata from `image_tensor_specs` utility by getting `metadata_extractor()` from our engine. A for loop over the `output_indices` check if the `outputmetadata` is empty (or more specifically, if `NormalizationOptions` in `outputmetadata` is empty) then use `InputMetadata`'s NormalizationOptions.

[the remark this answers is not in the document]
lu-wang-g: > 1. `InputMetaData` shouldn't ever be `NULL` I guess for this task, also `NormalizationOptions` shouldn't be `NULL` as well.

Those has been verified [here](https://github.com/tensorflow/tflite-support/blob/8c0a369197e09be72b4833235a2b04d21de3e09d/tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc#L137). You can share large amount of code with image_preprocessor (for both [initialization](https://github.com/tensorflow/tflite-support/blob/86d8bea9579c43ca597056a17f1a6dc5fd7e6b08/tensorflow_lite_support/cc/task/processor/image_preprocessor.cc#L73) and [process](https://github.com/tensorflow/tflite-support/blob/86d8bea9579c43ca597056a17f1a6dc5fd7e6b08/tensorflow_lite_support/cc/task/processor/image_preprocessor.cc#L106) ), that they both read metadata from a specified tensor (either input or output), and process FrameBuffer according to the metadata.

> 2. What if `OutputMetadata` exists but it doesn't have `NormalizationOptions`?
If the tensor is unit8, it's OK. But if the tensor is float, raise an error. See [here](https://github.com/tensorflow/tflite-support/blob/8c0a369197e09be72b4833235a2b04d21de3e09d/tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc#L204-L208).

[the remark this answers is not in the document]
jonpsy: > Those has been verified [here](https://github.com/tensorflow/tflite-support/blob/8c0a369197e09be72b4833235a2b04d21de3e09d/tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc#L137). 

In `BuildInputTensorSpecs`, when `InputMetaData` is `nullptr` it just skips over, and fills `normalization_options` as `abs::opt`. But then you're accessing it inside `ImagePreprocessor::Preprocess` [here](https://github.com/tensorflow/tflite-support/blob/86d8bea9579c43ca597056a17f1a6dc5fd7e6b08/tensorflow_lite_support/cc/task/processor/image_preprocessor.cc#L184-L186). Is this undefined behaviour?

jonpsy:
> You can have an end-2-end test using ImageTransformer to validate the implementation. Try the style transfer model here: https://www.tensorflow.org/lite/examples/style_transfer/overview#performance_benchmarks

@lu-wang-g  Correct me if I'm wrong, but this requires 2 input images, aka two tensors. But our entire library is, currently, based on single input image. So, what's the solution here?

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,181 @@
...
+  auto* input_metadata = engine_->metadata_extractor()->GetInputTensorMetadata(
+      input_indices.at(0));
+
+  // Use input metadata for normalization as fallback.
+  auto* processing_metadata =
+      output_metadata != nullptr ? output_metadata : input_metadata;
+
+  absl::optional<vision::NormalizationOptions> normalization_options;
+  ASSIGN_OR_RETURN(normalization_options,
+                   GetNormalizationOptionsIfAny(*processing_metadata));
```
jonpsy, line 90: Since this was wrapped inside `unknown namespace` inside `image_tensor_specs.cc` we might need to copy-paste the code here unfortunately.

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
```diff
@@ -0,0 +1,77 @@
...
+  //   the dimensions of the model input tensor,
+  // - convert it to the colorspace of the input tensor (i.e. RGB, which is the
+  //   only supported colorspace for now),
+  // - rotate it according to its `Orientation` so that inference is performed
+  //   on an "upright" image.
+  absl::StatusOr<vision::FrameBuffer> Postprocess();
+
+  const vision::NormalizationOptions& GetNormalizationOptions() {
+    return *options_.get();
+  };
```
lu-wang-g, line 60: Does it need to be exposed to public?

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
```diff
@@ -0,0 +1,77 @@
...
+
+ private:
+  using Postprocessor::Postprocessor;
+
+  // Whether the model features quantized inference type (QUANTIZED_UINT8). This
+  // is currently detected by checking if all output tensors data type is uint8.
+  bool has_uint8_outputs_;
+
+  absl::Status Init(std::unique_ptr<vision::NormalizationOptions> options);
```
lu-wang-g, line 69: Is "absl::Status Init(std::unique_ptr<vision::NormalizationOptions> options);" implemented anywhere?

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
lu-wang-g: Please share as much code as possible between`ImagePostprocessor::Init` and `ImageTensorSpecs::BuildInputImageTensorSpecs`. You can put a todo and implement it in a follow up PR.

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
```diff
@@ -0,0 +1,77 @@
...
+  // - resize it (with bilinear interpolation, aspect-ratio *not* Postserved) to
+  //   the dimensions of the model input tensor,
+  // - convert it to the colorspace of the input tensor (i.e. RGB, which is the
+  //   only supported colorspace for now),
+  // - rotate it according to its `Orientation` so that inference is performed
+  //   on an "upright" image.
+  absl::StatusOr<vision::FrameBuffer> Postprocess();
+
+  const vision::NormalizationOptions& GetNormalizationOptions() {
+    return *options_.get();
```
lu-wang-g, line 59: Is options_ a class member? Where is it initialized?

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
jonpsy: Now that we're extracting normalization options from metadata, we needn't store `options_` separately. I'll delete it, is that okay?

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
```diff
@@ -0,0 +1,77 @@
...
+//    - image input of size `[batch x height x width x channels]`.
+//    - batch inference is not supported (`batch` is required to be 1).
+//    - only RGB inputs are supported (`channels` is required to be 3).
+//    - if type is kTfLiteFloat32, NormalizationOptions are required to be
+//      attached to the metadata for input normalization.
+class ImagePostprocessor : public Postprocessor {
+ public:
+  static tflite::support::StatusOr<std::unique_ptr<ImagePostprocessor>>
+  Create(core::TfLiteEngine* engine,
+         const std::initializer_list<int> output_indices.
```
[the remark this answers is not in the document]
jonpsy: Can't because `Postprocessor` class stores `std::vector<int> output_indices_` as state variable. I've continued the same pattern for `input_indices`. We can just access `output_indices_.at(0)`. Sounds good?

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
jonpsy: No. Should be `Init(const std::vector<int>& input_indices)` instead.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
jonpsy: Just so we're clear, do you mean copy code from `BuildInputImageTensorSpecs` when you said "share".

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
jonpsy: It should store the `NormalizationOptions` state. Updated the PR.

jonpsy:
@lu-wang-g  Looks like there were some changes in the API? I've updated the PR accordingly.

lu-wang-g:
Thanks Sai for updating the code! Yes, some internal API is changed which could affect your code here. I'll take a look at the changes on Friday (catching up a ddl today, and team off-site event tomorrow. Sorry about the delay!).

jonpsy:
I'm getting worse PSNR score than the one in the [colab notebook](https://colab.research.google.com/github/tensorflow/hub/blob/master/examples/colab/image_enhancing.ipynb). Either my PSNR implementation is wrong or I haven't taken images correctly. I'll fix this.

Another concern is, the `ImageTransformer` fails to `CreateFromOptions` when only `InputMetadata` is provided. Not sure why, I'll investigate this further.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
lu-wang-g: "Share code" means `ImagePostprocessor::Init` and `ImageTensorSpecs::BuildInputImageTensorSpecs` use the same piece of code to do processing or validation.

jonpsy:
@lu-wang-g Hi, in [this](https://github.com/tensorflow/hub/blob/master/examples/colab/image_enhancing.ipynb) notebook. Why this happens?

```python
def preprocess_image(image_path):
.. 
..
  hr_size = (tf.convert_to_tensor(hr_image.shape[:-1]) // 4) * 4
  hr_image = tf.image.crop_to_bounding_box(hr_image, 0, 0, hr_size[0], hr_size[1])
  hr_image = tf.cast(hr_image, tf.float32)
```

It is converting my (50 x 50) image into (48 x 48) image.

lu-wang-g:
> @lu-wang-g Hi, in [this](https://github.com/tensorflow/hub/blob/master/examples/colab/image_enhancing.ipynb) notebook. Why this happens?
> 
> ```python
> def preprocess_image(image_path):
> .. 
> ..
>   hr_size = (tf.convert_to_tensor(hr_image.shape[:-1]) // 4) * 4
>   hr_image = tf.image.crop_to_bounding_box(hr_image, 0, 0, hr_size[0], hr_size[1])
>   hr_image = tf.cast(hr_image, tf.float32)
> ```
> 
> It is converting my (50 x 50) image into (48 x 48) image. The inference engine is pretty strict on (50 x 50) as input.

This is the TF model, which is different from the TFLite model. I think here, it tries to align the width/height to the multiples of 4. You can ignore this in the TFLite implementation.

jonpsy:
@lu-wang-g  Reading the same file using `tf.io.decode_jpg` and `DecodeImage` is giving different values.


image


image

They're off by 1-2 , but since I'm comparing PSNR these changes are effecting a lot. What do you think is the cause?

lu-wang-g:
> @lu-wang-g Reading the same file using `tf.io.decode_jpg` and `DecodeImage` is giving different values.
> 
> image
> 
> image
> 
> They're off by 1-2 , but since I'm comparing PSNR these changes are effecting a lot. What do you think is the cause?

That's a known issue. It is due to different Jepg reader algorithms used in tf.io.decode_jpg and DecodeImage. For testing purpose in Task library, you can use `DecodeImage` to parse your image.

jonpsy:
Okay, but in our test I wanted to test against a known PSNR value, with a known set of images. Which is why I loaded the same image on colab (as in the figure above) and our testing model. But since they're reading different value, PSNR value come off very different. 

So, how do I perform PSNR test? Or, should I perform PSNR test?

tensorflow_lite_support/cc/test/task/vision/image_transformer_test.cc
```diff
@@ -0,0 +1,146 @@
...
+  // Zero MSE means no noise is present in the signal.
+  double psnr = mse == 0 ? 100.0 : 20 * std::log10(maxPixel / std::sqrt(mse));
+
+  return psnr;
+}
+
+// Use a bi-cubically downsampled image as input to the model and compare
+// the model output with the original image. Expected PSNR taken from:
+// https://colab.research.google.com/github/tensorflow/hub/blob/master/examples/colab/image_enhancing.ipynb.
+TEST_F(SuperResolutionTest, PSNRTest) {
```
jonpsy, line 116: @lu-wang-g , continuing above comment. i was talkin about this test.

lu-wang-g: How much is your PSNR value? Is your output image clear than the input image, like is it enhanced?

jonpsy: It wasn't particularly bad, around ~25. I haven't inspected the output image. How do I convert FrameBuffer to jpeg?

jonpsy:
After doing EncodeToPng(), the output png is something like this.

husky_error

Possible reasons could be:

-  **Corrupted model**: To verify this, I used ipynb notebook and fed image to same model.  Which means this is not the case.
- **Incorrect postprocessing**: To verify this, I used `ofstream` to dumped the output of `FrameBuffer result_or.value()` to `.txt` file and read it in ipynb. Then, I performed RMSE loss between `image_transformer` output vs python inference with the same tflite model, the loss was around ~2.222 which is not bad. Which means this is not the case either.


This confirms the _error happens during writing to png_. Have a look at the [notebook](https://colab.research.google.com/drive/1silhAiURa2wBkOAUAqA_XHFVK8l11JT1?usp=sharing
) I created.

**Model output from python inference**
husky_python

**Model output from reading FrameBuffer output via txt file**
husky_cpp

Looks almost same right? That means our framework works properly, but storing to png is causing this weird behaviour. I'm not sure why though, maybe because input image dims is not divisible by 4? Let me know what you think @lu-wang-g

lu-wang-g:
> o

It is probably due to misusage of `EncodeToPng`, i.e. width and height are reversed. If you already have a working output image (as you pasted in the thread), just use that one in test. Good to confirm that the pipeline is good.

jonpsy:
> > o
> 
> It is probably due to misusage of `EncodeToPng`, i.e. width and height are reversed.

But then height and width have the same value (200), so not sure what's the reason. Let's discuss this over #718 since this PR isn't to blame.

>  If you already have a working output image (as you pasted in the thread), just use that one in test. Good to confirm that the pipeline is good.

Yeah, I'll use the image I got from dumping & displaying in ipynb. This should seal the deal with this PR.

jonpsy:
@lu-wang-g  I'm closing this PR & opening dividing into two as discussed.

PR #702, opened 2021-09-28
Integration of Error Reporting in C Image Classifier

priankakariatyml:
1. Added a method in c/common_utils to create errors originating in C layer.
2. Integrated error reporting in C Image Classifier.
3. Added error testing in C Image Classifier  tests.
4. Added checks for score and label of first category in classify tests.


tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
```diff
@@ -42,51 +42,88 @@ constexpr char kMobileNetQuantizedWithMetadata[] =
...
 
 class ImageClassifierFromOptionsTest : public tflite_shims::testing::Test {};
 
+TEST_F(ImageClassifierFromOptionsTest, FailsWithNullOptionsAndError) {
+  TfLiteSupportError* error = nullptr;
+  TfLiteImageClassifier* image_classifier =
+      TfLiteImageClassifierFromOptions(nullptr, &error);
+
+  EXPECT_EQ(image_classifier, nullptr);
+  if (image_classifier) TfLiteImageClassifierDelete(image_classifier);
```
lu-wang-g, line 57: nit: line 57 is redundant. image_classifier should just be a nullptr because of line 56.

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
```diff
@@ -42,51 +42,88 @@ constexpr char kMobileNetQuantizedWithMetadata[] =
...
+
+  ASSERT_NE(error, nullptr);
+  EXPECT_EQ(error->code, kInvalidArgumentError);
+  EXPECT_NE(error->message, nullptr);
+  EXPECT_NE(strstr(error->message, "Expected non null options"), nullptr);
```
lu-wang-g, line 62: Can you use something like:
EXPECT_THAT(error->message, HasSubstr("Expected non null options"));?

tensorflow_lite_support/ios/BUILD
```diff
@@ -66,9 +67,11 @@ tflite_ios_framework(
         ":base_options.h",
         ":bounding_box.h",
         ":classification_options.h",
+        ":common.h",
         ":classification_result.h",
         ":frame_buffer.h",
         ":image_classifier.h",
+
```
lu-wang-g, line 74: nit: remove line 74.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -167,8 +174,17 @@ TfLiteClassificationResult* GetClassificationResultCStruct(
...
+    ::tflite::support::CreateTfLiteSupportError(
+        kInvalidArgumentError, "Expected non null image classifier.", error);
+    return nullptr;
+  }
+
+  if (frame_buffer == nullptr) {
+    ::tflite::support::CreateTfLiteSupportError(
+        kInvalidArgumentError, "Expected non null frame buffer.", error);
     return nullptr;
   }
```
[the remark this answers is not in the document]
priankakariatyml: We can. But right now `frame_buffer` param of `CreateCppFrameBuffer ` is being passed by reference. So we have to check for `null` before the call. If we can change that to pass the` frame_buffer` as a pointer by value, we can check for `nullptr` in `CreateCppFrameBuffer ` and create a Status with appropriate message and return it.  This will enable us to remove the extra null check in the image_classifer for frame_buffer. Should I change the param to a pointer?

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -103,11 +102,19 @@ TfLiteImageClassifierOptions TfLiteImageClassifierOptionsCreate() {
...
+     ::tflite::support::CreateTfLiteSupportError(
+        kInvalidArgumentError, "Expected non null options.", error);
+     return nullptr;
+  }
+
   std::unique_ptr<ImageClassifierOptionsCpp> cpp_options =
       CreateImageClassifierCppOptionsFromCOptions(options);
 
   if (cpp_options == nullptr) {
+    ::tflite::support::CreateTfLiteSupportError(
```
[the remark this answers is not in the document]
priankakariatyml: Fixed

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -103,11 +102,19 @@ TfLiteImageClassifierOptions TfLiteImageClassifierOptionsCreate() {
...
-    const TfLiteImageClassifierOptions* options) {
+    const TfLiteImageClassifierOptions* options, TfLiteSupportError** error) {
+  if (options == nullptr) {
+     ::tflite::support::CreateTfLiteSupportError(
+        kInvalidArgumentError, "Expected non null options.", error);
+     return nullptr;
+  }
+
   std::unique_ptr<ImageClassifierOptionsCpp> cpp_options =
       CreateImageClassifierCppOptionsFromCOptions(options);
```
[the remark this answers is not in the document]
priankakariatyml: Changed return type to StatusOr<ImageClassifierOptionsCpp>

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
priankakariatyml: But `EXPECT_EQ` will not stop execution on failure. EXPECT_EQ fails when `image_classifier` is not null. Logically, it will have to be deleted after that right? Hence the null check too.
But I do agree that in reality, the delete will never be executed in this particular test as the image classifier would not get created unless the C or C++ layer makes the mistake of creating some default options with a valid ExternalFile if options parameter is null. Chances of that happening are very less.
But I still put it in for the sake of code correctness. LMK if it needs to be removed.

lu-wang-g: I see. I would change `EXPECT_EQ(image_classifier, nullptr);` => `ASSERT_EQ(image_classifier, nullptr);`. But it's up to you.

tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -167,8 +175,25 @@ TfLiteClassificationResult* GetClassificationResultCStruct(
...
+  }
+
+  if (frame_buffer == nullptr) {
+    ::tflite::support::CreateTfLiteSupportError(
+        kInvalidArgumentError, "Expected non null frame buffer.", error);
+    return nullptr;
+  }
+
+  StatusOr<std::unique_ptr<FrameBufferCpp>> cpp_frame_buffer_status =
+      ::tflite::task::vision::CreateCppFrameBuffer(*frame_buffer);
```
lu-wang-g, line 193: I agree to change CreateCppFrameBuffer and allow it to load a pointer instead of a reference. frame_buffer is created as a pointer;  it's more natural to pass it into CreateCppFrameBuffer directly.

lu-wang-g (commented):
Looks great! Once you updated  the two minor issues, we can check in the code.

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: Fixed

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
priankakariatyml: But if this image classifier case fails but error is still non null, it won't be deleted if we change to `ASSERT` right, since error checks and cleanup are after this `EXPECT`

lu-wang-g (approved):

tensorflow_lite_support/c/test/task/vision/image_classifier_test.cc
lu-wang-g: No, it won't. That's an alternative if you don't want to delete line 57, but  I'm OK with either way.

lu-wang-g (approved):

PR #704, opened 2021-10-06
Added Common Base Option, Error Codes and Classification Option Header Files for Objective C

priankakariatyml:
Added build and header files of Objective C Framework for the following:

1. Base options
2. Classification Options
3. Common Error Codes

### DIFF
--- a/tensorflow_lite_support/ios/BUILD
+++ b/tensorflow_lite_support/ios/BUILD
@@ -79,3 +79,14 @@ tflite_ios_framework(
         "//tensorflow_lite_support/c/task/vision:image_classifier",
     ],
 )
+
+objc_library(
+    name = "TFLCommon",
+    hdrs = [
+        "Sources/TFLCommon.h",
+    ],
+   visibility = [
+       "//tensorflow_lite_support:__subpackages__"
+    ],
+    module_name = "TFLCommon",
+)
--- a/tensorflow_lite_support/ios/Sources/TFLCommon.h
+++ b/tensorflow_lite_support/ios/Sources/TFLCommon.h
@@ -1,0 +1,211 @@
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+ 
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ 
+ http://www.apache.org/licenses/LICENSE-2.0
+ 
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ ==============================================================================*/
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * @enum TFLSupportErrorCode
+ * This enum specifies  error codes for TensorFlow Lite Task Library.
+ * It maintains a 1:1 mapping to TfLiteSupportErrorCode of C libray.
+ */
+typedef NS_ENUM(NSUInteger, TFLSupportErrorCode) {
+  
+  /** Unspecified error. */
+  TFLSupportErrorCodeUnspecifiedError = 1,
+  
+  /** Invalid argument specified. */
+  TFLSupportErrorCodeInvalidArgumentError = 2,
+  
+  /** Invalid FlatBuffer file or buffer specified. */
+  TFLSupportErrorCodeInvalidFlatBufferError = 3,
+  
+  /** Model contains a builtin op that isn't supported by the OpResolver or
+   delegates.  */
+  TFLSupportErrorCodeUnsupportedBuiltinOpError = 4,
+  
+  /** Model contains a custom op that isn't supported by the OpResolver or
+   * delegates. */
+  TFLSupportErrorCodeUnsupportedCustomOpError = 5,
+  
+  /** File I/O error codes. */
+  
+  /** No such file.  */
+  TFLSupportErrorCodeFileNotFoundError = 100,
+  
+  /** Permission issue. */
+  TFLSupportErrorCodeFilePermissionDeniedError,
+  
+  /** I/O error when reading file. */
+  TFLSupportErrorCodeFileReadError,
+  
+  /** I/O error when mmap-ing file. */
+  TFLSupportErrorCodeFileMmapError,
+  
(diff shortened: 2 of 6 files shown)


tensorflow_lite_support/ios/BUILD
```diff
@@ -79,3 +79,14 @@ tflite_ios_framework(
         "//tensorflow_lite_support/c/task/vision:image_classifier",
     ],
 )
+
+objc_library(
+    name = "TFLCommon",
+    hdrs = [
+        "Sources/TFLCommon.h",
```
lu-wang-g, line 86: What is the common code hierarchy of obj-c? I'm not familiar with obj-c, but wondering what the code style convention is. 

The current pattern is:
ios/
 - task/
   -  core/
      - Sources/ 
   - text/
     - api/
     - nlclassifier/
       - Sources/
       - Tests/
     - qa/
       - Sources/
       - Tests/
   - vison/
- Sources/ 
- text/
   - tokenizers/
     - Sources/
     - Tests/
 - utils/
 
Should we make Sources/ a root directory under ios/? For example:
ios/
  - Sources/
    - task/
      - core/
      - text/
      - vision/
    - TFLCommon.h
  - Tests

tensorflow_lite_support/ios/Sources/TFLCommon.h
```diff
@@ -0,0 +1,211 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ ==============================================================================*/
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * @enum TFLSupportErrorCode
```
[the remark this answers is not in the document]
priankakariatyml: AFAIK, ARC won't work on C structs. Hence we have to free C structs we use within our Obj C framework. If we reuse C structs, ObjC Client code will also have to take care of freeing them once allocated inspite off enabling ARC. Plus syntax for initialising, pointer reference etc for C data structures is non native to Obj C. You will still have to use the C syntax. iOS devs are not familiar with these. Hence we are hiding all these by way of creating native ObjC types over them. I have never called C funcs from Swift, so I am not sure if the same holds true in case of Swift client code as well.
 Enum `TfLiteSupportErrorCode` probably can be reused. But we loose the capability of defining things like `NS_SWIFT_NAME` for it. Also I have removed a lot of documentation that is specific to only C users of the library and some values that were defined for easy looping through the enum which are not needed in the ObjC library. Documentation maybe confusing if it's reused. I can investigate the enum reusability if need be.

tensorflow_lite_support/ios/BUILD
priankakariatyml: There is no standard specified in ObjC AFAIK. A lot of popular frameworks don't use the `sources` folder at all. They just group them in a hierarchy meaningfully. But most of the iOS frameworks from Google maintain a similar hierarchy. I just followed whatever was followed for the text library.  If you look at [tensorflow lite objc](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/lite/objc), they have two folders `sources` and `apis`. `Sources` are all the `.mm` files and internal headers. `Apis` are public headers. Task library omits api folder and puts everything in `Sources`. We can follow the structure you mentioned if it doesn't interfere with any Google standards. Won't matter make a difference from the iOS POV.  @khanhlvg Do you have more info about this?

khanhlvg: I think Lu's suggestion on the folder structure is more scalable. The only concern is that would it require significant changes on the build script for the current Text library?

```
ios/
  Sources/
    task/
       core/
       text/
       vision/
    TFLCommon.h
Tests
```

@lu-wang-g Do you know if Prianka can make the folder restructure or do we need to do it internally?

tensorflow_lite_support/ios/Sources/TFLCommon.h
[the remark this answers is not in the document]
khanhlvg: +1 on redefining the error code as ObjC's enum and structs as ObjC objects as it'll be easier for Swift interop. The API will be very hard for use for iOS developers if we reuse the C structs.

tensorflow_lite_support/ios/task/core/Sources/TFLBaseOptions.h
```diff
@@ -0,0 +1,72 @@
+//
```
khanhlvg, line 1: Please replace all headers with the TensorFlow's Apache License header.

tensorflow_lite_support/ios/BUILD
lu-wang-g: Let's loop in Youngsoek and see if it makes sense to reorganize the code structure and if it will cause backward incompatibility issue. This PR can be merged first.

lu-wang-g (approved):

tensorflow_lite_support/ios/BUILD
priankakariatyml: I can proceed with this hierarchy. There is also an apis folder in [ios/task/text](https://github.com/tensorflow/tflite-support/tree/master/tensorflow_lite_support/ios/task/text/apis) with umbrella header and module map. Where would these files go in the new structure? apis don't go inside sources, right? So will there be another apis folder like

priankakariatyml: But we will have to create a hierarchy for task in apis as well in that case right?
For eg:,
ios/
 apis/
    task/
      text/
        TFLTask.h
        framework.modulemap
      vision/
        TFLVision.h
        framework.modulemap
 Sources/
    task/
      text/
      vision/

lu-wang-g: I think so, there's another folder, apis/, to host the header files.

@priankakariatyml how about we start to follow the new structure for vision tasks. If you have extra bandwidth, you can help us to refactor the text tasks' code. Otherwise, I'll handle it when I have chance.  

Seems like the code is better to be organized according to each iOS package, i.e. for text / vision / audio, respectively. So the new structure could be something like this:
ios/
  Sources/   <= General source code for tflite_support. Doesn't have to be used in Task library.
     TFLCommon.h
  apis/ 
  task/
     core/
        Sources/  <= should this be "sources/"?  It's less common to capitalize directory name in Google's code 
        apis/
     text/
        Sources/ 
        apis/
     vision/
        Sources/ 
        apis/

What do you think? Please feel free to let me know. Thanks!

priankakariatyml: This looks good to me as well since there is an apis folder too. I can reorganise text and update it in the next PR.  So if I understand correctly, all public headers go into respective `api` folders. `.m` and `internal headers` go into `sources`. Am I right?

PR #706, opened 2021-10-14
Added Header Files and Rearranged Folder Structure for Vision

priankakariat:
Added header files
1. ios/task/processor/TFLClassificationResult.h
2. ios/TFLCommonUtil.h

Rearranged public headers into apis and internal headers into sources for vision

### DIFF
--- a/tensorflow_lite_support/cc/task/audio/utils/BUILD
+++ b/tensorflow_lite_support/cc/task/audio/utils/BUILD
@@ -1,5 +1,4 @@
-# Description:
-# wav_io targets.
+# Description: some utilitiy function for audio processing.
 
 package(
     default_visibility = [
--- a/tensorflow_lite_support/cc/task/audio/utils/wav_io.cc
+++ b/tensorflow_lite_support/cc/task/audio/utils/wav_io.cc
@@ -16,7 +16,7 @@ limitations under the License.
 // Functions to write audio in WAV format.
 // This file is forked from `tensorflow/core/lib/wav/wav_io.cc`.
 
-#include "tensorflow_lite_support/examples/task/audio/desktop/wav/wav_io.h"
+#include "tensorflow_lite_support/cc/task/audio/utils/wav_io.h"
 
 #include <math.h>
 #include <string.h>
--- a/tensorflow_lite_support/cc/task/audio/utils/wav_io.h
+++ b/tensorflow_lite_support/cc/task/audio/utils/wav_io.h
@@ -16,8 +16,9 @@ limitations under the License.
 // Functions to write audio in WAV format.
 // This file is forked from `tensorflow/core/lib/wav/wav_io.h`.
 
-#ifndef TENSORFLOW_LITE_SUPPORT_EXAMPLES_TASK_AUDIO_DESKTOP_WAV_WAV_IO_H_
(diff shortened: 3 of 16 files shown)


tensorflow_lite_support/ios/sources/TFLCommonUtils.h
```diff
@@ -0,0 +1,63 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#import <Foundation/Foundation.h>
+#import <TensorFlowLiteTaskVisionC/TensorFlowLiteTaskVisionC.h>
```
lu-wang-g, line 17: Is this needed?

tensorflow_lite_support/ios/sources/TFLCommonUtils.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param memSize size of memory to be allocated
+ * @param error Pointer to the memory location where errors if any should be saved. If `nil`, no
+ * error will be saved.
+ *
+ * @return Pointer to the allocated block of memory on successfull allocation. nil in case as error
+ * is encountered because of invalid memSize. If failure is due to any other reason, method
+ * terminates program execution.
+ */
++ (void *)mallocOrErrorWithSize:(size_t)memSize error:(NSError **)error;
```
lu-wang-g, line 60: Should this be  `mallocOrErrorWithSize` or `mallocErrorWithSize`?

tensorflow_lite_support/ios/task/processor/apis/TFLClassificationOptions.h
```diff
@@ -12,7 +12,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
  ==============================================================================*/
-
+
```
lu-wang-g, line 15: nit: remove the empty space.

tensorflow_lite_support/ios/task/processor/apis/TFLClassificationResult.h
```diff
@@ -0,0 +1,86 @@
...
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/** Encapsulates information about a class in the classification results. */
+@interface TFLCategory : NSObject
+
+/** Display name of the class. */
+@property(nonatomic, strong) NSString *displayName;
```
khanhlvg, line 24: As this is an entity class, I think we should use `copy` is better than `strong` to make sure that the values aren't changed unintentionally. Same for other properties of object type.

tensorflow_lite_support/ios/sources/TFLCommonUtils.h
khanhlvg: Thanks for the clarification. SGTM.

lu-wang-g: What is "TensorFlowLiteTaskVisionC.h"? It's a little wired to have TFLCommonUtils.h depends on Task Vision C API.

PR #720, opened 2021-12-03
Image postprocessor

jonpsy:
Continuation of #679. @lu-wang-g

### DIFF
--- a/tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
+++ b/tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
@@ -79,14 +79,17 @@ GetNormalizationOptionsIfAny(const TensorMetadata& tensor_metadata) {
 tflite::support::StatusOr<std::unique_ptr<ImagePostprocessor>>
 ImagePostprocessor::Create(core::TfLiteEngine* engine,
                            const std::initializer_list<int> output_indices,
-                           const std::initializer_list<int> input_indices) { 
-  ASSIGN_OR_RETURN(auto processor, Processor::Create<ImagePostprocessor>(/* num_expected_tensors = */ 1, engine, output_indices, /* requires_metadata = */ false));
+                           const std::initializer_list<int> input_indices) {
+  ASSIGN_OR_RETURN(auto processor,
+                   Processor::Create<ImagePostprocessor>(
+                       /* num_expected_tensors = */ 1, engine, output_indices,
+                       /* requires_metadata = */ false));
 
   RETURN_IF_ERROR(processor->Init(input_indices));
   return processor;
 }
 
-absl::Status ImagePostprocessor::Init(const std::vector<int>& input_indices) {
+absl::Status ImagePostprocessor::Init(const std::vector<int> &input_indices) {
   if (core::TfLiteEngine::OutputCount(engine_->interpreter()) != 1) {
     return tflite::support::CreateStatusWithPayload(
         absl::StatusCode::kInvalidArgument,
@@ -97,13 +100,15 @@ absl::Status ImagePostprocessor::Init(const std::vector<int>& input_indices) {
         tflite::support::TfLiteSupportStatus::kInvalidNumOutputTensorsError);
   }
 
-  if (GetTensor()->type != kTfLiteUInt8 && GetTensor()->type != kTfLiteFloat32) {
+  if (GetTensor()->type != kTfLiteUInt8 &&
+      GetTensor()->type != kTfLiteFloat32) {
     return tflite::support::CreateStatusWithPayload(
         absl::StatusCode::kInvalidArgument,
         absl::StrFormat("Type mismatch for output tensor %s. Requested one "
                         "of these types: "
                         "kTfLiteUint8/kTfLiteFloat32, got %s.",
-                        GetTensor()->name, TfLiteTypeGetName(GetTensor()->type)),
+                        GetTensor()->name,
+                        TfLiteTypeGetName(GetTensor()->type)),
         tflite::support::TfLiteSupportStatus::kInvalidOutputTensorTypeError);
   }
 
@@ -112,8 +117,9 @@ absl::Status ImagePostprocessor::Init(const std::vector<int>& input_indices) {
         absl::StatusCode::kInvalidArgument,
         absl::StrCat("The input tensor should have dimensions 1 x height x "
                      "width x 3. Got ",
-                     GetTensor()->dims->data[0], " x ", GetTensor()->dims->data[1],
-                     " x ", GetTensor()->dims->data[2], " x ",
+                     GetTensor()->dims->data[0], " x ",
+                     GetTensor()->dims->data[1], " x ",
+                     GetTensor()->dims->data[2], " x ",
                      GetTensor()->dims->data[3], "."),
         tflite::support::TfLiteSupportStatus::
             kInvalidInputTensorDimensionsError);
@@ -123,12 +129,15 @@ absl::Status ImagePostprocessor::Init(const std::vector<int>& input_indices) {
   const tflite::TensorMetadata* output_metadata =
       engine_->metadata_extractor()->GetOutputTensorMetadata(
           tensor_indices_.at(0));
-  const tflite::TensorMetadata* input_metadata = engine_->metadata_extractor()->GetInputTensorMetadata(
-      input_indices.at(0));
+  const tflite::TensorMetadata* input_metadata =
+      engine_->metadata_extractor()->GetInputTensorMetadata(
+          input_indices.at(0));
 
   // Use input metadata for normalization as fallback.
   const tflite::TensorMetadata* processing_metadata =
-      GetNormalizationOptionsIfAny(*output_metadata).value().has_value() ? output_metadata : input_metadata;
+      GetNormalizationOptionsIfAny(*output_metadata).value().has_value()
+          ? output_metadata
+          : input_metadata;
 
   absl::optional<vision::NormalizationOptions> normalization_options;
   ASSIGN_OR_RETURN(normalization_options,
@@ -180,7 +189,8 @@ absl::StatusOr<vision::FrameBuffer> ImagePostprocessor::Postprocess() {
     postprocessed_data.insert(postprocessed_data.begin(), &output_data[0],
                               &output_data[output_byte_size / sizeof(uint8)]);
   } else {  // Denormalize to [0, 255] range.
-    if (GetTensor()->bytes / sizeof(float) != output_byte_size / sizeof(uint8)) {
+    if (GetTensor()->bytes / sizeof(float) !=
+        output_byte_size / sizeof(uint8)) {
       return tflite::support::CreateStatusWithPayload(
           absl::StatusCode::kInternal,
           "Size mismatch or unsupported padding bytes between pixel data "
@@ -205,7 +215,7 @@ absl::StatusOr<vision::FrameBuffer> ImagePostprocessor::Postprocess() {
       for (size_t i = 0; i < output_byte_size / sizeof(uint8);
            ++i, ++denormalized_output_data, ++output_data) {
         *denormalized_output_data = static_cast<uint8>(std::round(std::min(
-            255.f, 
+            255.f,
             std::max(0.f, (*output_data) * norm_options.std_values[i % 3] +
                               norm_options.mean_values[i % 3]))));
       }
@@ -219,7 +229,7 @@ absl::StatusOr<vision::FrameBuffer> ImagePostprocessor::Postprocess() {
       vision::FrameBuffer::Create({postprocessed_plane}, to_buffer_dimension,
                                   vision::FrameBuffer::Format::kRGB,
                                   vision::FrameBuffer::Orientation::kTopLeft);
-  return *postprocessed_frame_buffer.get();;
+  return *postprocessed_frame_buffer.get();
 }
 
 }  // namespace processor
--- a/tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
+++ b/tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
@@ -1,0 +1,227 @@
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "tensorflow_lite_support/cc/task/processor/image_postprocessor.h"
+
+namespace tflite {
+namespace task {
+namespace processor {
+
+namespace {
+
+using ::absl::StatusCode;
+using ::tflite::metadata::ModelMetadataExtractor;
+using ::tflite::support::CreateStatusWithPayload;
+using ::tflite::support::StatusOr;
+using ::tflite::support::TfLiteSupportStatus;
+
+StatusOr<absl::optional<vision::NormalizationOptions>>
+GetNormalizationOptionsIfAny(const TensorMetadata& tensor_metadata) {
+  ASSIGN_OR_RETURN(
+      const tflite::ProcessUnit* normalization_process_unit,
+      ModelMetadataExtractor::FindFirstProcessUnit(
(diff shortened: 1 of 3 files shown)


tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,237 @@
...
+
+  // Use input metadata for normalization as fallback.
+  const tflite::TensorMetadata* processing_metadata =
+      GetNormalizationOptionsIfAny(*output_metadata).value().has_value()
+          ? output_metadata
+          : input_metadata;
+
+  absl::optional<vision::NormalizationOptions> normalization_options;
+  ASSIGN_OR_RETURN(normalization_options,
+                   GetNormalizationOptionsIfAny(*processing_metadata));
```
lu-wang-g, line 144: Simplify this by something like:
```
ASSIGN_OR_RETURN(auto normalization_options, GetNormalizationOptionsIfAny(*GetTensorMetadata()));

if (!normalization_options.has_value() && input_index>-1) {
  ASSIGN_OR_RETURN(normalization_options, GetNormalizationOptionsIfAny(*input_metadata));
}
```

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,237 @@
...
+  ASSIGN_OR_RETURN(auto processor,
+                   Processor::Create<ImagePostprocessor>(
+                       /* num_expected_tensors = */ 1, engine, output_indices,
+                       /* requires_metadata = */ false));
+
+  RETURN_IF_ERROR(processor->Init(input_indices));
+  return processor;
+}
+
+absl::Status ImagePostprocessor::Init(const std::vector<int> &input_indices) {
```
lu-wang-g, line 92: Still, this Init function is largely the same as `ImageTensorSpecs::BuildInputImageTensorSpecs`. We should update `BuildInputImageTensorSpecs` and reuse the logic between `ImagePostprocessor::Init` and `ImagePreprocessor::Init` as much as possible.

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
```diff
@@ -0,0 +1,70 @@
...
+//    - only RGB inputs are supported (`channels` is required to be 3).
+//    - if type is kTfLiteFloat32, NormalizationOptions are required to be
+//      attached to the metadata for output de-normalization. Uses input metadata
+//      as fallback in case output metadata isn't provided.
+class ImagePostprocessor : public Postprocessor {
+ public:
+  static tflite::support::StatusOr<std::unique_ptr<ImagePostprocessor>>
+  Create(core::TfLiteEngine* engine,
+         const std::initializer_list<int> output_indices,
+         const std::initializer_list<int> input_indices);
```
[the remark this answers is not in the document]
jonpsy: Yes, I think we decided that we should stick with single input/output model.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
jonpsy: Why re-assign twice? I suppose you meant

```c++
  ASSIGN_OR_RETURN(auto normalization_options, GetNormalizationOptionsIfAny(*GetTensorMetadata()));

  if (GetTensor()->type == kTfLiteFloat32) {
    if (!normalization_options.has_value() && input_index > -1) {
     ....
```
No?

jonpsy: Nice one! 👍

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
jonpsy: Understood, instead of passing the tensor, we pass a boolean switch to handle cases like [these](https://github.com/tensorflow/tflite-support/blob/f5dadc83b36d863700f080ca45877f8b3b3d7f47/tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc#L140-L141) & rename to ~BuildInputImageTensorSpecs~ =>  BuildImageTensorSpecs
```c++
StatusOr<ImageTensorSpecs> BuildImageTensorSpecs(
    const TfLiteEngine::Interpreter& interpreter,
    const tflite::metadata::ModelMetadataExtractor& metadata_extractor,
    const bool is_input) { // Notice the boolean
```

similarly for ~GetInputTensorMetadataIfAny~ => GetTensorMetadataIfAny [here](https://github.com/tensorflow/tflite-support/blob/f5dadc83b36d863700f080ca45877f8b3b3d7f47/tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc#L42).

Such that we could switch getting InputCount vs OutputCount, InputTensor access vs OutputTensor access easily. How's the plan?

tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc
```diff
@@ -132,55 +141,67 @@ StatusOr<absl::optional<NormalizationOptions>> GetNormalizationOptionsIfAny(
...
+        "Models are assumed to have a single output.",
+        TfLiteSupportStatus::kInvalidNumOutputTensorsError);
   }
 
-  // Input-related specifications.
-  const TfLiteTensor* input_tensor = TfLiteEngine::GetInput(&interpreter, 0);
-  if (input_tensor->dims->size != 4) {
+  // Input/output-related specifications.
+  const TfLiteTensor* tensor = is_input
+                                   ? TfLiteEngine::GetInput(&interpreter, 0)
```
lu-wang-g, line 178: Again, you need to pass TfLiteTensor into BuildImageTensorSpecs, because it's not always the first Input or output tensor.

tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc
```diff
@@ -132,55 +141,67 @@ StatusOr<absl::optional<NormalizationOptions>> GetNormalizationOptionsIfAny(
...
+  if (is_input && TfLiteEngine::InputCount(&interpreter) != 1) {
     return CreateStatusWithPayload(
         StatusCode::kInvalidArgument,
         "Models are assumed to have a single input.",
         TfLiteSupportStatus::kInvalidNumInputTensorsError);
+  } else if (!is_input && TfLiteEngine::OutputCount(&interpreter) != 1) {
+    return CreateStatusWithPayload(
+        StatusCode::kInvalidArgument,
+        "Models are assumed to have a single output.",
+        TfLiteSupportStatus::kInvalidNumOutputTensorsError);
```
lu-wang-g, line 173: These check should be removed.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,129 @@
...
+      auto output_specs,
+      vision::BuildImageTensorSpecs(*engine_->interpreter(),
+                                    *engine_->metadata_extractor(), false));
+  options_ = std::make_unique<vision::NormalizationOptions>(
+      output_specs.normalization_options.value());
+  return absl::OkStatus();
+}
+
+absl::StatusOr<vision::FrameBuffer> ImagePostprocessor::Postprocess() {
+  has_uint8_output_ = GetTensor()->type == kTfLiteUInt8;
```
lu-wang-g, line 63: Move "has_uint8_output_ = GetTensor()->type == kTfLiteUInt8;" to Init()

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,129 @@
...
+  const int kRgbPixelBytes = 3;
```
lu-wang-g, line 64: Create a constexpr, such as `constexpr int kRgbPixelBytes = 3;', and put it in the anonymous namespace with other `using` statements.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,129 @@
...
+    if (GetTensor()->bytes != output_byte_size) {
+      return tflite::support::CreateStatusWithPayload(
+          absl::StatusCode::kInternal,
+          "Size mismatch or unsupported padding bytes between pixel data "
+          "and output tensor.");
+    }
+    const uint8* output_data =
+        core::AssertAndReturnTypedTensor<uint8>(GetTensor()).value();
+    postprocessed_data.insert(postprocessed_data.begin(), &output_data[0],
+                              &output_data[output_byte_size / sizeof(uint8)]);
```
lu-wang-g, line 82: Use PopulateVector: https://github.com/tensorflow/tflite-support/blob/b5ffda7b6c4f0ebb82f2e1dd1a2697211ad89b9c/tensorflow_lite_support/cc/task/core/task_utils.h#L132

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,129 @@
...
+      GetBufferByteSize(to_buffer_dimension, vision::FrameBuffer::Format::kRGB);
+  std::vector<uint8> postprocessed_data(output_byte_size / sizeof(uint8), 0);
+
+  if (has_uint8_output_) {  // No denormalization required.
+    if (GetTensor()->bytes != output_byte_size) {
+      return tflite::support::CreateStatusWithPayload(
+          absl::StatusCode::kInternal,
+          "Size mismatch or unsupported padding bytes between pixel data "
+          "and output tensor.");
+    }
```
lu-wang-g, line 78: This size check is redundant, because both GetTensor()->bytes and output_byte_size are calculated from output tensor shape. They are from the same ground truth.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,129 @@
...
+        core::AssertAndReturnTypedTensor<uint8>(GetTensor()).value();
+    postprocessed_data.insert(postprocessed_data.begin(), &output_data[0],
+                              &output_data[output_byte_size / sizeof(uint8)]);
+  } else {  // Denormalize to [0, 255] range.
+    if (GetTensor()->bytes / sizeof(float) !=
+        output_byte_size / sizeof(uint8)) {
+      return tflite::support::CreateStatusWithPayload(
+          absl::StatusCode::kInternal,
+          "Size mismatch or unsupported padding bytes between pixel data "
+          "and output tensor.");
```
lu-wang-g, line 89: Same here, redundant check.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,129 @@
...
+    }
+    const uint8* output_data =
+        core::AssertAndReturnTypedTensor<uint8>(GetTensor()).value();
+    postprocessed_data.insert(postprocessed_data.begin(), &output_data[0],
+                              &output_data[output_byte_size / sizeof(uint8)]);
+  } else {  // Denormalize to [0, 255] range.
```
lu-wang-g, line 83: Need to check if the data type is float or not. There can be other data types, such as int8 and int, which won't be supported here.

jonpsy: That's checked already in `valid_types` check before.

tensorflow_lite_support/cc/task/processor/image_postprocessor.h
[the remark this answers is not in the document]
lu-wang-g: Correct, you can turn `output_index` into an initlializer_list in `Create`. I have it in my TODO list to update the other processors' API to use a single integer whenever possible.

tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc
```diff
@@ -39,29 +39,38 @@ using ::tflite::support::StatusOr;
...
+  } else if (is_input && metadata_extractor.GetInputTensorCount() != 1) {
     return CreateStatusWithPayload(
         StatusCode::kInvalidArgument,
         "Models are assumed to have a single input TensorMetadata.",
         TfLiteSupportStatus::kInvalidNumInputTensorsError);
+  } else if (!is_input && metadata_extractor.GetOutputTensorCount() != 1) {
+    return CreateStatusWithPayload(
+        StatusCode::kInvalidArgument,
+        "Models are assumed to have a single output TensorMetadata.",
+        TfLiteSupportStatus::kInvalidNumOutputTensorsError);
```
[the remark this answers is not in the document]
lu-wang-g: `metadata_extractor.GetOutputTensorCount()` returns the output tensor count of the tflite model, which can be greater than one. For example, a model may takes two input image tensors. You can bind an ImagePreprocessor to each of the input image tensor, respectively, but not both input tensors to one ImagePreprocessor (that's what we've agreed in ImagePre(/Post)processor). However, the code will fail here, because `metadata_extractor.GetOutputTensorCount()` returns 2. 

This is a bug in our existing code, it's not discovered before, because all models we supported so far have only one input tensor of image. But we need to correct the logic.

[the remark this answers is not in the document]
lu-wang-g: And if you pass in `TensorMetadata`, it will be garnered that the number of tensor is 1 for the ImagePostprocessor.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
lu-wang-g: Checking the type data here is less error prone is later a new type is added in Init(), but forgot to update it here.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
```diff
@@ -0,0 +1,129 @@
...
+          "Size mismatch or unsupported padding bytes between pixel data "
+          "and output tensor.");
+    }
+
+    uint8* denormalized_output_data = postprocessed_data.data();
+    const float* output_data =
+        core::AssertAndReturnTypedTensor<float>(GetTensor()).value();
+    const auto norm_options = GetNormalizationOptions();
+
+    if (norm_options.num_values == 1) {
```
lu-wang-g, line 97: If you don't have a testing float model yet, this is a good one: 
https://github.com/margaretmz/Selfie2Anime-with-TFLite/blob/master/android/app/src/main/ml/selfie2anime.tflite
See the introduction of the model [here](https://github.com/margaretmz/Selfie2Anime-with-TFLite).

jonpsy:
Hey, do you mind if I store `input_index` in `ImagePostprocessor` class? The reason is, `input_index` is required in `Postprocess()` to extract fallback input metadata. As you said, we may use several postprocessor for a single model, hence we need to know appropriate `input_index` so that the correct fallback is chosen.

tensorflow_lite_support/cc/task/vision/utils/image_tensor_specs.cc
```diff
@@ -132,55 +141,67 @@ StatusOr<absl::optional<NormalizationOptions>> GetNormalizationOptionsIfAny(
...
-
-StatusOr<ImageTensorSpecs> BuildInputImageTensorSpecs(
+StatusOr<ImageTensorSpecs> BuildImageTensorSpecs(
     const TfLiteEngine::Interpreter& interpreter,
-    const tflite::metadata::ModelMetadataExtractor& metadata_extractor) {
+    const tflite::metadata::ModelMetadataExtractor& metadata_extractor,
+    const bool is_input) {
   ASSIGN_OR_RETURN(const TensorMetadata* metadata,
-                   GetInputTensorMetadataIfAny(metadata_extractor));
+                   GetTensorMetadataIfAny(metadata_extractor, is_input));
```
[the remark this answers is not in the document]
jonpsy: Okay but we need to use `GetTensorMetadataIfAny` for these checks:
```c++
  if (metadata_extractor.GetModelMetadata() == nullptr ||
      metadata_extractor.GetModelMetadata()->subgraph_metadata() == nullptr) 
```
and it requires `metadata_extractor`. So probably we should keep it?

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
jonpsy: @lu-wang-g  Okay, I will store `has_float32_output_` in `Init()` that should do the trick?

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
jonpsy: Okay, we decided against it.

tensorflow_lite_support/cc/task/processor/image_postprocessor.cc
jonpsy: The ESRGAN model works for both uint8 and float.

PR #721, opened 2021-12-03
[Part 2] Image Transformer 

jonpsy:
Continuation of #720


tensorflow_lite_support/cc/test/task/vision/image_transformer_test.cc
```diff
@@ -0,0 +1,145 @@
...
+  options.mutable_base_options()->mutable_model_file()->set_file_name(
+    JoinPath("./" /*test src dir*/, kTestDataDirectory,
+              kESRGANModelWithInputAndOutputMetaData));
+  SUPPORT_ASSERT_OK_AND_ASSIGN(std::unique_ptr<ImageTransformer> image_transformer,
+                      ImageTransformer::CreateFromOptions(options));
+
+  StatusOr<FrameBuffer> result_or =
+      image_transformer->Transform(*husky_downsampled_buffer);
+  SUPPORT_ASSERT_OK(result_or);
+  EXPECT_DOUBLE_EQ(PSNR(result_or.value(), *husky_original_buffer), 25.073790631326489);
```
jonpsy, line 137: Building on your idea of storing the "husky_test.png" (the output I got from dumping into file => reading in ipynb => saving as png).  I thought of using that "golden image" to compare with `result_or.value()`. The problem is, once again, `LoadImage` due to which the pixels are a bit off. So I think normal PSNR solution is the best stand-in solution atm.

PR #734, opened 2022-01-21
Added Ios object detection Swift Test

priankakariatyml:
1. Added Object Detection Swift Tests and Helper Files
2. Updated Image Classifier Swift Tests to use GMLImage initialiser from Helper Category


tensorflow_lite_support/ios/test/task/vision/object_detector/TFLObjectDetectorTests.m
```diff
@@ -104,21 +70,21 @@ - (void)testModelOptionsWithMaxResults {
...
-                  0.609375,                        // expectedFirstScore
-                  @"cat"                           // expectedFirstLabel
-  );
-  VerifyDetection(detectionResult.detections[2],
-                  CGRectMake(260, 394, 179, 209),  // expectedBoundingBox
-                  0.5625,                          // expectedFirstScore
-                  @"cat"                           // expectedFirstLabel
-  );
+
+  [TFLTestUtils verifyDetection:detectionResult.detections[0]
```
khanhlvg, line 74: Please don't use a util class for test like this. The feedback we got from internal review which led to the creation of the macros above for test code is that when a test fail, it will tell exactly which line it failed. It makes diagnosing issues easier. If you put all test code into a shared class, it will be harder to identify the cause because it only show that the shared class failed. 

Besides, unlike application code, Google prefers test code readability over reusability, so let's avoid sharing test code between ObjC and Swift and rewrite them so that each test is self-contained.

Please keep these test macros on this ObjC. For the Swift test, you can write them as a verifyDetection method.

priankakariatyml: Got it. Fixed

khanhlvg (approved):

PR #738, opened 2022-01-27
Added Python Image Classifier APIs and tests

kinaryml:


tensorflow_lite_support/python/task/processor/processor_options.py
```diff
@@ -31,3 +31,33 @@ class EmbeddingOptions:
...
+  # (if any). Results below this value are rejected. It is tempting to
+  # assume that the classification threshold should always be 0.5, but
+  # thresholds are problem-dependent, and are therefore values that you
+  # must tune.
+  score_threshold: Optional[float] = None
+
+  # Optional whitelist of class names. If non-empty, classifications whose
+  # class name is not in this set will be filtered out. Duplicate or unknown
+  # class names are ignored. Mutually exclusive with class_name_blacklist.
+  class_name_whitelist: List[str] = None
```
khanhlvg, line 58: Rename to the more respectful terms "allow_list" and "deny_list".

tensorflow_lite_support/python/task/processor/proto/class_pb2.py
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Class options protobuf."""
+
+from tensorflow_lite_support.cc.task.vision.proto import class_pb2
+
+Class = class_pb2.Class
```
khanhlvg, line 18: @lu-wang-g @mbrenon It looks like `Class` is called `Category` in both the C (iOS) and Java API. Should we do the same for the Python API?
https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/c/task/processor/category.cc
https://www.tensorflow.org/lite/api_docs/java/org/tensorflow/lite/support/label/Category

tensorflow_lite_support/python/task/vision/image_classifier.py
```diff
@@ -0,0 +1,136 @@
...
+  # Builds the initial proto_options.
+  proto_options = image_classifier_options_pb2.ClassifierOptions()
+
+  # Updates values from base_options.
+  proto_options.model_file_with_metadata.file_name = options.base_options.model_file
+  proto_options.num_threads = options.base_options.num_threads
+  if options.base_options.use_coral:
+    proto_options.compute_settings.tflite_settings.delegate = configuration_pb2.Delegate.EDGETPU_CORAL
+
+  # Updates values from classifier_options.
```
khanhlvg, line 50: As these dataclass to proto conversion logic will be reused in other tasks, please put them on a shared module (e.g `python/task/core/utils.py`) for better code reuse.

tensorflow_lite_support/python/task/processor/processor_options.py
```diff
@@ -31,3 +31,33 @@ class EmbeddingOptions:
   # therefore any dimension is guaranteed to have a value in [-1.0, 1.0]. Use
   # the l2_normalize option if this is not the case.
   quantize: Optional[bool] = None
+
+
+@dataclasses.dataclass
+class ClassificationOptions:
+  """Options for classification processor."""
+  # The locale to use for display names specified through the TFLite Model
```
khanhlvg, line 39: please use docstring instead of comment so that it'll show up in API docs later. For example:

```python
display_names_locale: Optional[str] = None
"""Metadata, if any. Defaults to English."""
```

tensorflow_lite_support/python/task/vision/image_classifier_test.py
```diff
@@ -0,0 +1,535 @@
...
+      base_options = task_options.BaseOptions(model_file=model_file, num_threads=-2)
+      options = image_classifier.ImageClassifierOptions(base_options=base_options)
+      image_classifier.ImageClassifier.create_from_options(options)
+
+  @parameterized.parameters(
+    (_MODEL_FLOAT, ['foo'], ['bar']),
+    (_MODEL_QUANTIZED, ['foo'], ['bar']),
+    (_MODEL_AUTOML, ['foo'], ['bar']),
+  )
+  def test_combined_whitelist_and_blacklist(self, model_name,
```
khanhlvg, line 111: rename to allowlist and denylist.

khanhlvg (changes requested):

tensorflow_lite_support/python/task/vision/image_classifier_test.py
```diff
@@ -0,0 +1,535 @@
...
+
+  @parameterized.parameters(
+    (_MODEL_FLOAT, 3, False),
+    (_MODEL_QUANTIZED, 3, False),
+    (_MODEL_AUTOML, 3, False),
+    (_MODEL_FLOAT, 3, True),
+    (_MODEL_QUANTIZED, 3, True),
+    (_MODEL_AUTOML, 3, True),
+  )
+  def test_classify_model(self, model_name, max_results, with_bounding_box):
```
khanhlvg, line 140: @kinaryml 
Please split to `test_classify_model` (without bounding box) and `test_classify_model_with_bounding_box` for better readability.

@lu-wang-g @mbrenon @ziyeqinghan 
I think it's fine to test with just one TFLite model file (e.g. only with the float model) as it's the responsibility of the underlying C++ layer to ensure that Task Library works correctly with both float and quantized model. WDYT?

tensorflow_lite_support/python/task/vision/image_classifier_test.py
```diff
@@ -0,0 +1,535 @@
...
+            head_index: 0
+          }
+          """)
+      )
+
+
+  @parameterized.parameters(
+    (_MODEL_FLOAT, None, 0.5, False),
+  )
+  def test_score_threshold_option(self, model_name,  max_results,
```
khanhlvg, line 405: You don't need to test a combination of options. Let's keep the test code simple by testing one option by one. The Python wrapper only pass the parameters from the Python layer to C++ layer, so it isn't the responsibility of the Python layer to worry about combining options.

tensorflow_lite_support/python/task/vision/image_classifier_test.py
```diff
@@ -0,0 +1,535 @@
...
+              score: 0.01171875
+              class_name: "dandelion"
+            }
+            head_index: 0
+          }
+          """)
+      )
+
+
+  @parameterized.parameters(
```
khanhlvg, line 402: If there's one test case, don't use parameterized test and just write the test code normally.

tensorflow_lite_support/python/task/vision/image_classifier_test.py
kinaryml: Updated the file to use two separate test cases (classification w/ and w/o the bounding box).

tensorflow_lite_support/python/task/vision/image_classifier_test.py
kinaryml: Simplified the test cases

tensorflow_lite_support/python/task/processor/processor_options.py
```diff
@@ -11,23 +11,69 @@
...
 
 @dataclasses.dataclass
 class EmbeddingOptions:
   """Options for embedding processor."""
-  # Whether to normalize the returned feature vector with L2 norm. Use this
-  # option only if the model does not already contain a native L2_NORMALIZATION
-  # TF Lite Op. In most cases, this is already the case and L2 norm is thus
-  # achieved through TF Lite inference.
+
+  """
```
khanhlvg, line 24: Sorry, I misunderstood the requirements for Python docstring. Each attributes should be documented in the class definition as below.

https://github.com/tensorflow/tensorflow/blob/d248815ff16901941ca617b020a26739f0cce230/tensorflow/lite/python/lite.py#L188

tensorflow_lite_support/python/task/processor/proto/class_pb2.py
mbrenon: Yes, let's maybe go for consistency across wrappers where possible and call that `Category` here too. The `Class` naming is historical and we can hardly get rid of it in the C++ API, but with hindsight tends to cause a lot of problems because it's usually a reserved keyword...

tensorflow_lite_support/python/task/processor/processor_options.py
```diff
@@ -11,23 +11,69 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Options to configure Embedding APIs."""
+"""Options to configure Vision APIs."""
```
mbrenon, line 14: Let's be even more generic here and go for "Options to configure processors"? We're going to have to change it anyway in the future with other types of APIs.

tensorflow_lite_support/python/task/vision/core/utils.py
```diff
@@ -0,0 +1,28 @@
...
+"""Utils for Vision APIs."""
+
+from tensorflow_lite_support.python.task.core.proto import configuration_pb2
+
+
+def build_proto_base_options(proto_options, options):
+  """Builds initial proto options values from the base options."""
+
+  proto_options.model_file_with_metadata.file_name = options.base_options.model_file
+  proto_options.num_threads = options.base_options.num_threads
```
mbrenon, line 23: I'm not sure I understand why you're doing this (filling deprecated options from their newer counterpart). Cc @khanhlvg in case I'm missing something.

kinaryml: @mbrenon Uh the image embedder python module seemed to be using the same so I figured I'd use that. I'll look into it and make sure it uses the new counterpart.

mbrenon: Got it. Indeed, the image embedder options don't use `BaseOptions` at all so this logic is required there, but it's not required here.

kinaryml: I've fixed it and made sure it uses the newer BaseOptions and ComputeSettings counterparts.

tensorflow_lite_support/python/task/vision/image_classifier_test.py
```diff
@@ -0,0 +1,499 @@
...
+from tensorflow_lite_support.python.task.processor.proto import bounding_box_pb2
+from tensorflow_lite_support.python.task.vision import image_classifier
+from tensorflow_lite_support.python.task.vision.core import tensor_image
+from tensorflow_lite_support.python.test import test_util
+
+import unittest
+import textwrap
+
+
+_MODEL_FLOAT = "mobilenet_v2_1.0_224.tflite"
```
khanhlvg, line 29: @lu-wang-g @mbrenon @ziyeqinghan
I think it's fine to test with just one TFLite model file (e.g. only with the float model) as it's the responsibility of the underlying C++ layer to ensure that Task Library works correctly with both float and quantized model. WDYT?

tensorflow_lite_support/python/task/processor/proto/BUILD
```diff
@@ -22,3 +22,24 @@ py_library(
     srcs = ["bounding_box_pb2.py"],
     deps = ["//tensorflow_lite_support/cc/task/vision/proto:bounding_box_py_pb2"],
 )
+
+py_library(
```
mbrenon, line 26: For consistency with the latest changes on ImageEmbedder: this one (image_classifier_options) should live under python/task/vision/pybinds. The rationale between this is that this proto is never exposed to the user directly.

tensorflow_lite_support/python/task/processor/proto/image_classifier_options_pb2.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2022 The TensorFlow Authors. All Rights Reserved.
```
mbrenon, line 1: See above comment: please move to python/task/vision/pybinds

tensorflow_lite_support/python/task/vision/BUILD
```diff
@@ -42,3 +42,45 @@ py_test(
...
+        "//tensorflow_lite_support/python/task/processor/proto:classifications_pb2",
+        "//tensorflow_lite_support/python/task/processor/proto:image_classifier_options_pb2",
+        "//tensorflow_lite_support/python/task/vision/core:tensor_image",
+        "//tensorflow_lite_support/python/task/vision/core:utils",
+        "//tensorflow_lite_support/python/task/vision/core/pybinds:image_utils",
+        "//tensorflow_lite_support/python/task/vision/pybinds:image_classifier",
+    ],
+)
+
+py_test(
```
mbrenon, line 66: Please move tests to python/test/task/vision

tensorflow_lite_support/python/task/vision/core/utils.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2021 The TensorFlow Authors. All Rights Reserved.
```
mbrenon, line 1: Please hold on to that one: an equivalent will be checked in (as python/task/core/task_utils.py) very soon, you will be able to use that one instead.

tensorflow_lite_support/python/task/vision/image_classifier.py
```diff
@@ -0,0 +1,134 @@
...
+from tensorflow_lite_support.python.task.processor.proto import bounding_box_pb2
+from tensorflow_lite_support.python.task.processor.proto import classifications_pb2
+from tensorflow_lite_support.python.task.processor.proto import image_classifier_options_pb2
+from tensorflow_lite_support.python.task.vision.core import tensor_image
+from tensorflow_lite_support.python.task.vision.core.pybinds import image_utils
+from tensorflow_lite_support.python.task.vision.pybinds import image_classifier as _image_classifier
+
+from tensorflow_lite_support.python.task.vision.core.utils import build_proto_base_options
+
+
```
mbrenon, line 30: Please adopt the same conventions as https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/python/task/vision/image_embedder.py, i.e:

_ProtoImageClassifierOptions = image_classifier_options_pb2.ImageClassifierOptions
_CppImageClassifier = _pywrap_image_classifier.ImageClassifier

tensorflow_lite_support/python/task/vision/image_classifier_test.py
mbrenon: Agreed.

tensorflow_lite_support/python/task/vision/pybinds/BUILD
```diff
@@ -23,3 +23,20 @@ pybind_extension(
         "@pybind11_protobuf//pybind11_protobuf:native_proto_caster",
     ],
 )
+
+pybind_extension(
+    name = "image_classifier",
+    srcs = [
+        "image_classifier.cc",
+    ],
+    module_name = "image_classifier",
```
mbrenon, line 32: Please adopt the same conventon as image embedder here, i.e. "_pywrap_image_classifier"

tensorflow_lite_support/python/task/vision/pybinds/image_classifier.cc
```diff
@@ -0,0 +1,94 @@
...
+      break;
+    default:
+      // This should not never happen because the eventual caller in
+      // interpreter.py should have passed a valid id here.
+      TFLITE_DCHECK(false);
+      return nullptr;
+  }
+  return resolver;
+}
+
```
mbrenon, line 58: Please remove (see image embedder)

tensorflow_lite_support/python/task/vision/pybinds/image_classifier.cc
```diff
@@ -0,0 +1,94 @@
...
+}  // namespace
+
+PYBIND11_MODULE(image_classifier, m) {
```
mbrenon, line 61: _pywrap_image_classifier

tensorflow_lite_support/python/task/vision/pybinds/image_classifier.cc
```diff
@@ -0,0 +1,94 @@
...
+  pybind11::google::ImportStatusModule();
+  pybind11_protobuf::ImportNativeProtoCasters();
+
+  py::class_<ImageClassifier>(m, "ImageClassifier")
+      .def_static(
+          "create_from_options",
+          [](const ImageClassifierOptions& options, int op_resolver_id) {
+            return ImageClassifier::CreateFromOptions(
+                options, get_resolver(op_resolver_id));
+          },
```
mbrenon, line 73: See image embedder for how to handle op_resolver being removed

mbrenon (changes requested):

tensorflow_lite_support/python/task/processor/proto/BUILD
kinaryml: @mbrenon I'll be closing this PR since there are major changes needed. I thought of raising this larger PR to get some early feedback FYI

PR #751, opened 2022-02-24
Added comprehensive python tests for Image Classifier Task Library API

kinaryml:


tensorflow_lite_support/python/test/task/vision/image_classifier_test.py
```diff
@@ -128,6 +164,129 @@ def test_classify_model(self, model_file_type, max_results,
...
+
+    # Loads image.
+    image = tensor_image.TensorImage.from_file(self.test_image_path)
+
+    # Classifies the input.
+    image_result = classifier.classify(image, bounding_box=None)
+    image_result_dict = json.loads(json_format.MessageToJson(image_result))
+
+    # Expected results.
+    expected_categories = [
```
khanhlvg, line 238: As you're testing the allowlist, you should assert if there's all class names in image_result belongs to the class_name_allowlist, rather than comparing it with a predefined result.

tensorflow_lite_support/python/test/task/vision/image_classifier_test.py
```diff
@@ -128,6 +164,129 @@ def test_classify_model(self, model_file_type, max_results,
...
+
+    # Loads image
+    image = tensor_image.TensorImage.from_file(self.test_image_path)
+
+    # Classifies the input.
+    image_result = classifier.classify(image, bounding_box=None)
+    image_result_dict = json.loads(json_format.MessageToJson(image_result))
+
+    # Expected results.
+    expected_categories = [
```
khanhlvg, line 264: Same as above. as you're testing the deny, you should assert if none of the class names in image_result belongs to the class_name_denylist, rather than comparing it with a predefined result.

tensorflow_lite_support/python/test/task/vision/image_classifier_test.py
```diff
@@ -128,6 +164,129 @@ def test_classify_model(self, model_file_type, max_results,
...
+        r"INVALID_ARGUMENT: `class_name_whitelist` and `class_name_blacklist` "
+        r"are mutually exclusive options. "
+        r"\[tflite::support::TfLiteSupportStatus='2'\]"):
+      base_options = _BaseOptions(model_file=_ExternalFile(file_name=self.model_path))
+      classification_options = classification_options_pb2.ClassificationOptions(
+        class_name_allowlist=['foo'], class_name_denylist=['bar'])
+      options = _ImageClassifierOptions(
+        base_options=base_options, classification_options=classification_options)
+      _ImageClassifier.create_from_options(options)
+
```
khanhlvg, line 290: please add tests for max_results and score_threshold. Similarly, these tests don't need to compare the model output with a predefined list, but assert if it's logically correct.

kinaryml: @khanhlvg 
- I've now changed them to assert if they function logically correct instead of comparing results with the ground truth. Similarly added a case to test the correctness of the `max_results` option too. 
- Also tidied up the code by adding global constants to be used for each test case.

khanhlvg (approved):

tensorflow_lite_support/python/test/task/vision/image_classifier_test.py
```diff
@@ -126,7 +168,138 @@ def test_classify_model(self, model_file_type, max_results,
...
+                       'Label "{0}" found but in deny list.'.format(label))
+
+  def test_combined_allowlist_and_denylist(self):
+    # Fails with combined allowlist and denylist
+    with self.assertRaisesRegex(
+        Exception,
+        r"INVALID_ARGUMENT: `class_name_whitelist` and `class_name_blacklist` "
+        r"are mutually exclusive options. "
+        r"\[tflite::support::TfLiteSupportStatus='2'\]"):
+      base_options = _BaseOptions(model_file=_ExternalFile(file_name=self.model_path))
```
ziyeqinghan, line 297: Please follow the google c++ style guide https://google.github.io/styleguide/cppguide.html#Line_Length: Each line of text in your code should be at most 80 characters long.

Please keep in mind to follow the google c++ style guide. Thanks! :)

kinaryml: @ziyeqinghan Thanks for the reference to the guide and the line-length limit used by Google! I'll hard wrap them to 80 from now on.

ziyeqinghan (approved):

PR #753, opened 2022-03-01
Implementation for the Python Object Detection Task Library 

kinaryml:


tensorflow_lite_support/python/test/task/vision/object_detector_test.py
```diff
@@ -0,0 +1,118 @@
...
+        bounding_box_pb2.BoundingBox(**bounding_box))
+      detection.classes.append(class_pb2.Category(**category))
+      expected_result.detections.append(detection)
+
+    expected_result_dict = json.loads(
+      json_format.MessageToJson(expected_result))
+
+    return expected_result_dict
+
+  def test_top1_result(self):
```
kinaryml, line 84: @khanhlvg I've added a test

khanhlvg: thanks!

ziyeqinghan (approved):

ziyeqinghan:
LGTM. Thanks!

tensorflow_lite_support/python/test/task/vision/object_detector_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Performs object detection on the input.
+    image_result = detector.detect(image)
+    image_result_dict = json.loads(json_format.MessageToJson(image_result))
+
+    # Builds test data.
+    expected_detections = [
+      (
+        {
+          'origin_x': 54, 'origin_y': 396, 'width': 393, 'height': 196
```
fergushenderson, line 102: Hi.  Firstly, thank you for your contributions!

Secondly, I have some feedback on this PR.
This is extremely brittle and fragile test code.
Please don't write tests like this in future.
TensorFlow Lite does not guarantee bit-exact results.
The output from an ML model may vary depending on the details of exactly how (and in what order) the intermediate floating point calculations were done.

Rather than testing for exact values for the bounding box coordinates,
it is better to just verify that the computed bounding box coordinates are
within a reasonable range (e.g. a few pixels) of the expected bounding box.

The code in this CL does use an _ACCEPTABLE_ERROR_RANGE = 0.000001 parameter
to do the comparison, but that error margin is much too small for comparing pixel
coordinate values -- an error of 1 part in a million is much less than one pixel.

kinaryml: Hi @fergushenderson. At the time of implementing the tests for these APIs I was merely following the C++ API and was expected to keep it consistent with those APIs. Those APIs happened to test the exact values within a reasonable range and I did too. So there’s that. Hope it clarifies everything.

kinaryml: https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/cc/test/task/vision/object_detector_test.cc

For your reference.

kinaryml: https://github.com/tensorflow/tflite-support/blob/fe9f90bf2a3fd15352b1ed1399937b1fc859c235/tensorflow_lite_support/cc/test/task/vision/object_detector_test.cc#L113

Mostly followed this at the time of development.

fergushenderson: I think all of the tests in object_detector_test.cc that test for exact bounding box values are using `PostprocessTest`.  Those tests aren't getting the bounding box data from TensorFlow Lite, instead they mock out the object detection code, supplying hard-coded values for the bounding box location data,

https://github.com/tensorflow/tflite-support/blob/fe9f90bf2a3fd15352b1ed1399937b1fc859c235/tensorflow_lite_support/cc/test/task/vision/object_detector_test.cc#L446

and are just testing the post-processing.  So in those tests, it is probably fine to test for exact bounding box values.

However, the fact that these tests mock out the object detection code is not immediately clear from the code of the tests, instead being far away in the FillAndGetOutputTensors method of the PostprocessTest class, so I can see why this distinction wasn't obvious! :)

fergushenderson:
I am sending a PR to address this in tensorflow_lite_support/python/test/task/vision/object_detector_test.py

tensorflow_lite_support/python/test/task/vision/object_detector_test.py
kinaryml: @fergushenderson As for the post processing test cases that holds true. However the C++ tests did indeed compare exact bounding box values after loading a real TF Lite model (with metadata) and a real image as you can see here.

https://github.com/tensorflow/tflite-support/blob/fe9f90bf2a3fd15352b1ed1399937b1fc859c235/tensorflow_lite_support/cc/test/task/vision/object_detector_test.cc#L327

This does use the same `ExpectApproximatelyEqual` to compare values FYI.
In that case, I believe the C++ tests need to be revised as well by using a margin for the pixel coordinates error to compare the actual and expected outputs.

kinaryml: The post processing tests aren’t involved in those Detect tests.

fergushenderson: > the C++ tests did indeed compare exact bounding box values after loading a real TF Lite model

OK, good point, I stand corrected.

> In that case, I believe the C++ tests need to be revised as well by using a margin for the pixel coordinates error to compare the actual and expected output

Agreed.  It looks like Alan Kelly <https://github.com/alankelly> has made a change for that already, but for some reason it hasn't propagated into github master yet.

PR #758, opened 2022-03-10
Added iOS Image Segmentation Implementation Files

priankakariatyml:
Added .m files and changed BUILD files for Image segmentation headers.
Added a fix in GMLImageUtils to fix incorrect reading of segmentation test image.
Added a single swift test for sanity check and modified GMLImage+Helpers to support the same. (Full implementation of the test and Objective C tests will be added in a later PR.)


tensorflow_lite_support/ios/test/task/vision/utils/sources/GMLImage+Helpers.m
```diff
@@ -28,4 +28,35 @@ + (GMLImage *)imageFromBundleWithClass:(Class)classObject
   return [[GMLImage alloc] initWithImage:image];
 }
 
+- (CVPixelBufferRef)grayScalePixelBufferFromUIImage {
```
khanhlvg, line 31: This is better as a category of UIImage rather than GMLImage. Could you move it to GMLImageUtils?

khanhlvg (changes requested):

tensorflow_lite_support/ios/test/task/vision/utils/sources/GMLImage+Helpers.m
priankakariatyml: For the time being I have removed this from the PR since it is needed for tests. I'll updated it in the follow up PR when the tests are added.

khanhlvg (approved):

tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
```diff
@@ -0,0 +1,80 @@
...
+
+@implementation TFLCategoryMask
+@synthesize mask;
+@synthesize width;
+@synthesize height;
+
+- (id)copyWithZone:(NSZone *)zone {
+  TFLCategoryMask *categoryMask = [[TFLCategoryMask alloc] init];
+
+  [categoryMask setWidth:self.width];
```
khanhlvg, line 25: @priankakariatyml Please use property instead of calling the setter directly. Please fix other places in the rest of the code.
```
categoryMask.width = self.width;
categoryMask.height = self.height;
```

tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
```diff
@@ -0,0 +1,80 @@
...
+  [categoryMask setHeight:self.height];
+  [categoryMask setMask:self.mask];
```
khanhlvg, line 27: mask is a pointer so please deep copy it instead of just copying the pointer.

tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
```diff
@@ -0,0 +1,80 @@
...
+
+@end
+
+@implementation TFLSegmentation
+@synthesize confidenceMasks;
+@synthesize categoryMask;
+@synthesize coloredLabels;
+
+- (void)dealloc {
+  if (self.confidenceMasks != NULL) {
```
khanhlvg, line 66: Please use nil instead of NULL in ObjC.

tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
```diff
@@ -0,0 +1,80 @@
...
+    for (int j = 0; j < self.confidenceMasks.count; ++j) {
```
khanhlvg, line 67: Instead of making the TFLSegmentation to own the lifecycle of the masks, you should move the deallocation logic to the dealloc method of `TFLConfidenceMask` and `TFLCategoryMask` accordingly. There's a good chance that users only care about the masks and will continue to use it but discard the TFLSegmentation instance immediately.

khanhlvg (changes requested):

tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
priankakariatyml: Fixed

tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
priankakariatyml: Fixed

khanhlvg:
Merged in https://github.com/tensorflow/tflite-support/commit/39e8d3212a9d42bf707d92edea2e8154388d0caa

PR #762, opened 2022-03-16
Task Library: Core Audio PyBinds and `AudioRecord`

kinaryml:


tensorflow_lite_support/python/task/audio/core/audio_record.py
```diff
@@ -0,0 +1,96 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""A module to record audio in a streaming basis."""
+import threading
+import numpy as np
+import sounddevice as sd
```
ziyeqinghan, line 17: Please  add unittest for this. May I ask when will we use this class?

ziyeqinghan (approved):

tensorflow_lite_support/python/task/audio/core/audio_record.py
khanhlvg: @kinaryml To test AudioRecord properly, I think you'll need to mock InputStream.

@ziyeqinghan AudioRecord is to create a wrapper around sounddevice to record audio from microphone in realtime, in the format that matches the requirement of the model. You can preview the API in the reference app.

https://github.com/tensorflow/examples/blob/7b53db0db388ab24b8957f77e702387dcf729dfd/lite/examples/sound_classification/raspberry_pi/classify.py#L53

kinaryml: @khanhlvg Any there any examples that I can refer to for mocking InputStream?

khanhlvg: Unfortunately there's no example but I summarized some directions into this issue.
https://github.com/tensorflow/tflite-support/issues/764

PR #766, opened 2022-03-23
Python TensorAudio implementation and tests

kinaryml:
- Added `TensorAudio` class
- Added tests for `TensorAudio`

### DIFF
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -99,26 +99,48 @@ class TensorAudio(object):
     data = record.read(self._buffer_size)
     self.load_from_array(data.astype(np.float32))
 
-  def load_from_array(self, src: np.ndarray) -> None:
-    """Loads audio data from a NumPy array.
+  def load_from_array(self,
+                      src: np.ndarray,
+                      offset: int = None,
+                      size: int = None) -> None:
+    """Loads the audio data from a NumPy array.
 
     Args:
-      src: A NumPy array contains the input audio.
-
+      src: A NumPy source array contains the input audio.
+      offset: An optional offset for loading a slice of the `src` array to the
+        buffer.
+      size: An optional size parameter denoting the number of samples to load
+        from the `src` array.
     Raises:
-      ValueError if the input audio is too large or if it contains an invalid
-      number of channels.
+      ValueError: Raised if the input array has an incorrect shape or if
+      offset + size exceeds the length of the `src` array.
     """
-    if len(src) != len(self._buffer):
-      raise ValueError(
-        f"Input audio contains an invalid number of samples. "
-        f"Expect {len(self._buffer)}.")
-    elif src.shape[1] != self._format.channels:
+    if src.shape[1] != self._format.channels:
       raise ValueError(
         f"Input audio contains an invalid number of channels. "
         f"Expect {self._format.channels}.")
 
-    self._buffer = src
+    if size is None:
+      size = len(src)
+    if offset is None:
+      offset = 0
+
+    if offset + size > len(src):
+      raise ValueError(
+        f"Index out of range. offset {offset} + size {size} should be <= "
+        f"src's length: {len(src)}")
+
+    if len(src) > len(self._buffer):
+      # Copy values from the source array to the internal buffer.
+      new_offset = offset + size - len(self._buffer)
+      new_size = len(self._buffer)
+      self._buffer = src[new_offset:new_offset + new_size]
+    else:
+      # Shift the internal buffer backward and add the incoming data to the end
+      # of the buffer.
+      shift = len(src)
+      self._buffer = np.roll(self._buffer, -shift, axis=0)
+      self._buffer[-shift:, :] = src
 
   @property
   def format(self) -> _CppAudioFormat:
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -57,8 +57,8 @@ class TensorAudio(object):
 
   @classmethod
   def create_from_wav_file(cls,
-                    file_name: str,
-                    buffer_size: int) -> "TensorAudio":
+                           file_name: str,
+                           buffer_size: int) -> "TensorAudio":
     """Creates `TensorAudio` object from the WAV file.
 
     Args:
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -101,7 +101,7 @@ class TensorAudio(object):
 
     # Load audio data from the AudioRecord instance.
     data = record.read(self._sample_count)
-    self.load_from_array(data)
+    self.load_from_array(data.astype(np.float32))
 
   def load_from_array(self, src: np.ndarray) -> None:
     """Loads audio data from a NumPy array.
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -29,7 +29,7 @@ class TensorAudio(object):
   def __init__(self,
                audio_format: _CppAudioFormat,
                buffer_size: int,
-               audio_data: np.ndarray = None,
+               buffer: np.ndarray = None,
                is_from_file: bool = False,
                ) -> None:
     """Initializes the `TensorAudio` object.
@@ -37,7 +37,7 @@ class TensorAudio(object):
     Args:
       audio_format: format of the audio.
       buffer_size: buffer size of the audio.
-      audio_data: contains raw audio data.
+      buffer: contains raw audio data.
       is_from_file: whether `audio_data` is loaded from the audio file.
     """
     self._format = audio_format
@@ -45,7 +45,7 @@ class TensorAudio(object):
     self._is_from_file = is_from_file
 
     if self._is_from_file:
-      self._buffer = audio_data
+      self._buffer = buffer
     else:
       self._buffer = np.zeros(
         [self._buffer_size, self._format.channels], dtype=np.float32)
@@ -77,7 +77,7 @@ class TensorAudio(object):
     audio = _LoadAudioBufferFromFile(
       file_name, buffer_size, np.zeros([buffer_size]))
     return cls(audio.audio_format, audio.buffer_size,
-               audio.float_buffer, is_from_file=True)
+               np.array(audio.float_buffer, copy=False), is_from_file=True)
 
   def load_from_audio_record(self, record: audio_record.AudioRecord) -> None:
     """Loads audio data from an AudioRecord instance.
@@ -139,8 +139,4 @@ class TensorAudio(object):
   @property
   def buffer(self) -> np.ndarray:
     """Gets the internal buffer."""
-    if self._is_from_file:
-      buffer = np.array(self._buffer, copy=False)
-    else:
-      buffer = self._buffer
-    return buffer
+    return self._buffer
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -28,28 +28,28 @@ class TensorAudio(object):
 
   def __init__(self,
                audio_format: _CppAudioFormat,
-               sample_count: int,
+               buffer_size: int,
                audio_data: _CppAudioBuffer = None,
                is_from_file: bool = False,
                ) -> None:
     """Initializes the `TensorAudio` object.
 
     Args:
-      audio_format: C++ AudioFormat object, format of the audio.
-      sample_count: int, number of samples in the audio.
-      audio_data: C++ AudioBuffer object, contains raw audio data, buffer size
+      audio_format: AudioFormat, format of the audio.
+      buffer_size: int, buffer size of the audio.
+      audio_data: AudioBuffer, contains raw audio data, buffer size
       and audio format info.
       is_from_file: boolean, whether `audio_data` is loaded from the audio file.
     """
     self._format = audio_format
-    self._sample_count = sample_count
+    self._buffer_size = buffer_size
     self._is_from_file = is_from_file
 
     if self._is_from_file:
-      self._data = audio_data
+      self._buffer = audio_data
     else:
       self._buffer = np.zeros(
-        [self._sample_count, self._format.channels], dtype=np.float32)
+        [self._buffer_size, self._format.channels], dtype=np.float32)
 
   def clear(self):
     """Clear the internal buffer and fill it with zeros."""
@@ -58,7 +58,7 @@ class TensorAudio(object):
   @classmethod
   def create_from_wav_file(cls,
                            file_name: str,
-                           buffer_size: int) -> "TensorAudio":
+                           buffer_size: int = None) -> "TensorAudio":
     """Creates `TensorAudio` object from the WAV file.
 
     Args:
@@ -69,13 +69,16 @@ class TensorAudio(object):
       `TensorAudio` object.
 
     Raises:
-      status.StatusNotOk if the audio file can't be decoded. Need to import
-        the module to catch this error: `from pybind11_abseil import status`,
-        see https://github.com/pybind/pybind11_abseil#abslstatusor.
+      status.StatusNotOk if the audio file can't be decoded.
     """
+    # TODO(b/220931229): Raise RuntimeError instead of status.StatusNotOk.
+    # Need to import the module to catch this error:
+    # `from pybind11_abseil import status`
+    # see https://github.com/pybind/pybind11_abseil#abslstatusor.
     audio = _LoadAudioBufferFromFile(
       file_name, buffer_size, np.zeros([buffer_size]))
-    return cls(audio.audio_format, audio.buffer_size, audio, is_from_file=True)
+    return cls(audio.audio_format, audio.buffer_size,
+               audio.float_buffer, is_from_file=True)
 
   def load_from_audio_record(self, record: audio_record.AudioRecord) -> None:
     """Loads audio data from an AudioRecord instance.
@@ -84,7 +87,7 @@ class TensorAudio(object):
     Raises:
       ValueError: Raised if the audio record's config is invalid.
     """
-    if record.buffer_size < self._sample_count:
+    if record.buffer_size < self._buffer_size:
       raise ValueError(
         "The audio record's buffer size cannot be smaller than the tensor "
         "audio's sample count.")
@@ -100,7 +103,7 @@ class TensorAudio(object):
         f"Expects {self._format.sample_rate}Hz.")
 
     # Load audio data from the AudioRecord instance.
-    data = record.read(self._sample_count)
+    data = record.read(self._buffer_size)
     self.load_from_array(data.astype(np.float32))
 
   def load_from_array(self, src: np.ndarray) -> None:
@@ -130,16 +133,15 @@ class TensorAudio(object):
     return self._format
 
   @property
-  def sample_count(self) -> int:
+  def buffer_size(self) -> int:
     """Gets the sample count of the audio."""
-    return self._sample_count
+    return self._buffer_size
 
   @property
-  def data(self) -> _CppAudioBuffer:
-    """Gets the C++ AudioBuffer object."""
+  def buffer(self) -> np.ndarray:
+    """Gets the internal buffer."""
     if self._is_from_file:
-      audio_data = self._data
+      audio_data = np.array(self._buffer, copy=False)
     else:
-      audio_data = _CppAudioBuffer(
-        self._buffer, self._sample_count, self._format)
+      audio_data = self._buffer
     return audio_data
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -18,7 +18,6 @@ import numpy as np
 from tensorflow_lite_support.python.task.audio.core.pybinds import _pywrap_audio_buffer
 from tensorflow_lite_support.python.task.audio.core import audio_record
 
-_CppAudioBuffer = _pywrap_audio_buffer.AudioBuffer
 _CppAudioFormat = _pywrap_audio_buffer.AudioFormat
 _LoadAudioBufferFromFile = _pywrap_audio_buffer.LoadAudioBufferFromFile
 
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -33,7 +33,6 @@ class TensorAudio(object):
     Args:
       audio_format: format of the audio.
       buffer_size: buffer size of the audio.
-      audio_data, optional: contains raw audio data.
     """
     self._format = audio_format
     self._buffer_size = buffer_size
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -28,26 +28,21 @@ class TensorAudio(object):
   def __init__(self,
                audio_format: _CppAudioFormat,
                buffer_size: int,
-               buffer: np.ndarray = None,
-               is_from_file: bool = False,
-               ) -> None:
+               audio_data: np.ndarray = None) -> None:
     """Initializes the `TensorAudio` object.
 
     Args:
       audio_format: format of the audio.
       buffer_size: buffer size of the audio.
-      buffer: contains raw audio data.
-      is_from_file: whether `audio_data` is loaded from the audio file.
+      audio_data, optional: contains raw audio data.
     """
     self._format = audio_format
     self._buffer_size = buffer_size
-    self._is_from_file = is_from_file
-
-    if self._is_from_file:
-      self._buffer = buffer
-    else:
-      self._buffer = np.zeros(
+    self._buffer = np.zeros(
         [self._buffer_size, self._format.channels], dtype=np.float32)
+
+    if audio_data is not None:
+      self.load_from_array(np.array(audio_data, copy=False))
 
   def clear(self):
     """Clear the internal buffer and fill it with zeros."""
@@ -75,8 +70,7 @@ class TensorAudio(object):
     # see https://github.com/pybind/pybind11_abseil#abslstatusor.
     audio = _LoadAudioBufferFromFile(
       file_name, buffer_size, np.zeros([buffer_size]))
-    return cls(audio.audio_format, audio.buffer_size,
-               np.array(audio.float_buffer, copy=False), is_from_file=True)
+    return cls(audio.audio_format, audio.buffer_size, audio.float_buffer)
 
   def load_from_audio_record(self, record: audio_record.AudioRecord) -> None:
     """Loads audio data from an AudioRecord instance.
--- a/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
+++ b/tensorflow_lite_support/python/task/audio/core/tensor_audio.py
@@ -53,7 +53,7 @@ class TensorAudio(object):
       file_name: WAV file name.
       buffer_size: Required input buffer size. The number of samples that the
       C++ AudioBuffer object can store. If the WAV file contains more samples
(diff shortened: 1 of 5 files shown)


tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,145 @@
...
+
+_CppAudioBuffer = _pywrap_audio_buffer.AudioBuffer
+_CppAudioFormat = _pywrap_audio_buffer.AudioFormat
+_LoadAudioBufferFromFile = _pywrap_audio_buffer.LoadAudioBufferFromFile
+
+
+class TensorAudio(object):
+  """A wrapper class to store the input audio."""
+
+  def __init__(self,
```
khanhlvg, line 29: Let's change the approach in the storage mode of TensorAudio a little bit:
* TensorAudio will only store data in a numpy array buffer. It's the responsibility of each task to convert TensorAudio into a `_pywrap_audio_buffer.AudioBuffer` to pass to the C++ implementation of each task.
* `create_from_wav_file` loads the audio data into the _pywrap_audio_buffer.AudioBuffer first, then convert to Numpy array and store it. You should use non-copy conversion like [this](https://github.com/tensorflow/tflite-support/blob/5490f096933fe0348d45c8ca8c955e1f158f8971/tensorflow_lite_support/python/task/vision/core/tensor_image.py#L96)
* Rename `sample_count` to `buffer_size` to align with the naming of the C++ layer.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,145 @@
...
+    else:
+      self._buffer = np.zeros(
+        [self._sample_count, self._format.channels], dtype=np.float32)
+
+  def clear(self):
+    """Clear the internal buffer and fill it with zeros."""
+    self._buffer.fill(0)
+
+  @classmethod
+  def create_from_wav_file(cls,
```
khanhlvg, line 59: * buffer_size should be optional. If it's not specified, buffer_size should be the full size of the WAV file. If buffer_size is specified, only load the buffer_size from the beginning of the WAV file. Please document this behavior in docstring.
* Please also update the C++ LoadAudioBufferFromFile function to follow this new behavior so that we have consistency between the two.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,145 @@
...
+    """Gets the audio format of the audio."""
+    return self._format
+
+  @property
+  def sample_count(self) -> int:
+    """Gets the sample count of the audio."""
+    return self._sample_count
+
+  @property
+  def data(self) -> _CppAudioBuffer:
```
khanhlvg, line 138: Rename this to buffer, and return the NumPy buffer.

tensorflow_lite_support/python/test/task/audio/core/tensor_audio_test.py
```diff
@@ -0,0 +1,165 @@
...
+    tensor = tensor_audio.TensorAudio.create_from_wav_file(
+      self.test_audio_path, _SAMPLE_COUNT)
+    tensor_audio_format = tensor.format
+
+    self.assertEqual(tensor_audio_format.channels, _CHANNELS)
+    self.assertEqual(tensor_audio_format.sample_rate, _SAMPLE_RATE)
+    self.assertEqual(tensor.sample_count, _SAMPLE_COUNT)
+    self.assertIsInstance(tensor.data, _CppAudioBuffer)
+
+  def test_load_from_array(self):
```
khanhlvg, line 52: Rename to `test_load_from_array_succeeds` to align with naming of failed test cases.

tensorflow_lite_support/python/test/task/audio/core/tensor_audio_test.py
```diff
@@ -0,0 +1,165 @@
...
+  def test_create_from_wav_file(self):
+    # Loads TensorAudio object from WAV file.
+    tensor = tensor_audio.TensorAudio.create_from_wav_file(
+      self.test_audio_path, _SAMPLE_COUNT)
+    tensor_audio_format = tensor.format
+
+    self.assertEqual(tensor_audio_format.channels, _CHANNELS)
+    self.assertEqual(tensor_audio_format.sample_rate, _SAMPLE_RATE)
+    self.assertEqual(tensor.sample_count, _SAMPLE_COUNT)
+    self.assertIsInstance(tensor.data, _CppAudioBuffer)
```
khanhlvg, line 50: Please assert some values in the buffer to ensure the WAV file is loaded correctly.

tensorflow_lite_support/python/test/task/audio/core/tensor_audio_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class TensorAudioTest(unittest.TestCase):
+  def setUp(self):
+    super().setUp()
+    self.test_audio_path = test_util.get_test_data_path('speech.wav')
+    self.test_tensor_audio = tensor_audio.TensorAudio(
+      _CppAudioFormat(_CHANNELS, _SAMPLE_RATE), _SAMPLE_COUNT)
+
+  def test_create_from_wav_file(self):
```
khanhlvg, line 41: Please add failed test cases, such as loading a file that doesn't exist. Also rename the success test case to test_create_from_wav_file_succeeds to align with failed test cases naming.

khanhlvg (changes requested):

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,145 @@
...
+               audio_format: _CppAudioFormat,
+               sample_count: int,
+               audio_data: _CppAudioBuffer = None,
+               is_from_file: bool = False,
+               ) -> None:
+    """Initializes the `TensorAudio` object.
+
+    Args:
+      audio_format: C++ AudioFormat object, format of the audio.
+      sample_count: int, number of samples in the audio.
```
khanhlvg, line 39: There's no need to specify type hint in docstring. It can be inferred directly from the Python code.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
kinaryml: @khanhlvg The example [here](https://github.com/tensorflow/tflite-support/blob/5490f096933fe0348d45c8ca8c955e1f158f8971/tensorflow_lite_support/examples/task/audio/desktop/audio_classifier_lib.cc#L77) seems to expect the buffer size as a required input for the LoadAudioBufferFromFile function. How do we determine the size of the buffer [here](https://github.com/khanhlvg/tflite-support/blob/45e15cf5973f3462095a00362cb544c16b54d5d9/tensorflow_lite_support/python/task/audio/core/tensor_audio.py#L79) in Python if the buffer size is unspecified?

khanhlvg: Please disregard this comment. I was thinking that we should provide an option to load the whole WAV file into the TensorAudio, but then I realized that Audio classifier still doesn't support classifying an arbitrary size audioBuffer yet, so let's keep it as how it's now.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,141 @@
...
+_CppAudioFormat = _pywrap_audio_buffer.AudioFormat
+_LoadAudioBufferFromFile = _pywrap_audio_buffer.LoadAudioBufferFromFile
+
+
+class TensorAudio(object):
+  """A wrapper class to store the input audio."""
+
+  def __init__(self,
+               audio_format: _CppAudioFormat,
+               buffer_size: int,
```
khanhlvg, line 30: Let's keep the init method simple. You don't need to pass the buffer here but use the load from array function. Besides, there's no use for the is_from_file parameter so let's remove it.

tensorflow_lite_support/python/test/task/audio/core/tensor_audio_test.py
```diff
@@ -0,0 +1,170 @@
...
+    # Loads TensorAudio object from WAV file.
+    tensor = tensor_audio.TensorAudio.create_from_wav_file(
+      self.test_audio_path, _BUFFER_SIZE)
+    tensor_audio_format = tensor.format
+
+    self.assertEqual(tensor_audio_format.channels, _CHANNELS)
+    self.assertEqual(tensor_audio_format.sample_rate, _SAMPLE_RATE)
+    self.assertEqual(tensor.buffer_size, _BUFFER_SIZE)
+    self.assertIsInstance(tensor.buffer, np.ndarray)
+    self.assertEqual(tensor.buffer[-1], -0.09640503)
```
khanhlvg, line 50: As you're asserting float number, you should use assertAlmostEqual

tensorflow_lite_support/python/test/task/audio/core/tensor_audio_test.py
```diff
@@ -0,0 +1,170 @@
...
+      callback_fn(dummy_data)
+    expected_data = np.concatenate(input_data[-2:])
+
+    # Load audio data into TensorAudio from the AudioRecord instance.
+    self.test_tensor_audio.load_from_audio_record(record)
+
+    # Assert read all data in the float buffer.
+    testing.assert_almost_equal(self.test_tensor_audio.buffer, expected_data)
+
+  def test_load_from_audio_record_fails_with_invalid_buffer_size(self):
```
khanhlvg, line 138: Sounddevice doesn't run on devices without a soundcard so you'll need to mock InputStream on all test cases that use AudioRecord.

kinaryml: @khanhlvg Alright, will fix it

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
kinaryml: Okay, got it.

tensorflow_lite_support/python/test/task/audio/core/tensor_audio_test.py
kinaryml: @khanhlvg Fixed it.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
kinaryml: @khanhlvg I've simplified the init method. Could you check it out?

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
khanhlvg: Can you remove audio_data from initialization params?

kinaryml: @khanhlvg How do we save the float buffer object with the `create_from_wav_file` method being a classmethod?

khanhlvg: Can you get the float buffer from the loaded AudioBuffer, then create a numpy array from it and call load_from_array after initializing the TensorAudio?

You can do something like this:
https://github.com/tensorflow/tflite-support/blob/5490f096933fe0348d45c8ca8c955e1f158f8971/tensorflow_lite_support/python/task/vision/core/tensor_image.py#L96

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    # TODO(b/220931229): Raise RuntimeError instead of status.StatusNotOk.
+    # Need to import the module to catch this error:
+    # `from pybind11_abseil import status`
+    # see https://github.com/pybind/pybind11_abseil#abslstatusor.
+    audio = _LoadAudioBufferFromFile(
+      file_name, buffer_size, np.zeros([buffer_size]))
+    tensor = TensorAudio(audio.audio_format, audio.buffer_size)
+    tensor.load_from_array(np.array(audio.float_buffer, copy=False))
+    return tensor
```
kinaryml, line 70: @khanhlvg I think this should do it.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,132 @@
...
+        [self._buffer_size, self._format.channels], dtype=np.float32)
+
+  def clear(self):
+    """Clear the internal buffer and fill it with zeros."""
+    self._buffer.fill(0)
+
+  @classmethod
+  def create_from_wav_file(cls,
+                           file_name: str,
+                           buffer_size: int) -> "TensorAudio":
```
ziyeqinghan, line 49: Is this value required? Can we have a default value for `buffer_size`?

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,132 @@
...
+    # Need to import the module to catch this error:
+    # `from pybind11_abseil import status`
+    # see https://github.com/pybind/pybind11_abseil#abslstatusor.
+    audio = _LoadAudioBufferFromFile(
+      file_name, buffer_size, np.zeros([buffer_size]))
+    tensor = TensorAudio(audio.audio_format, audio.buffer_size)
+    tensor.load_from_array(np.array(audio.float_buffer, copy=False))
+    return tensor
+
+  def load_from_audio_record(self, record: audio_record.AudioRecord) -> None:
```
ziyeqinghan, line 72: Why can't we use this as a classmethod and name it as create_from_audio_record?

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,132 @@
...
+    if record.sampling_rate != self._format.sample_rate:
+      raise ValueError(
+        f"The audio record's sampling rate doesn't match. "
+        f"Expects {self._format.sample_rate}Hz.")
+
+    # Load audio data from the AudioRecord instance.
+    data = record.read(self._buffer_size)
+    self.load_from_array(data.astype(np.float32))
+
+  def load_from_array(self, src: np.ndarray) -> None:
```
ziyeqinghan, line 98: ditto

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
khanhlvg: This is required because each audio model has a fixed input size, hence they can only take a TensorAudio of a fixed size as the input.

@kinaryml Let's clarify better in the docstring that this method only loads buffer_size samples at the beginning of the file, same as the docstring of the underlying C++ implementation.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
khanhlvg: The logic here is that:
* TensorAudio buffer size must be the same as the AudioClassifer/AudioEmbedder model's input size. 
* Therefore, TensorAudio instances should be created by the AudioClassifer/AudioEmbedder object. The input size can be transparent to the model user, and they only need to *load* the data into the TensorAudio object without worrying about the input data size.
* AudioRecord object doesn't know anything about the TFLite model. Therefore, creating a TensorAudio from an AudioRecord instance isn't useful because it may not the compatible with the TFLite model that the user wants to use.

This behavior is similar to that of the Java's TensorAudio, but there's only one difference is that the Java's TensorAudio uses a ring buffer, while this Python's TensorAudio only implements a simple buffer.

@kinaryml After re-reading the [Java's TensorAudio implementation](AudioClassifer/AudioEmbedder), I think we should implement the feature of the TensorAudio as a ring buffer. In particular:
* load_from_array should be able to load an arbitrary size array, similar to the `public void load(float[] newData, int offset, int size)` of the Java implementation. 
  * If the size of the array is larger than the TensorAudio buffer size, only load the most recent data from the array.
  * If the size of the array is smaller than the TensorAudio buffer size, roll the internal buffer backward and add the new data to the end of the buffer (similar to the AudioRecord buffer implementation).
* load_from_array should support `offset` and `size` as optional parameters.

@ziyeqinghan @lu-wang-g @mbrenon WDYT?

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
kinaryml: @khanhlvg Fixed it.

khanhlvg: Let's change it a bit for better clarification:
* Rename buffer_size to sample_count.
* sample_count: The number of samples to read from the WAV file. This value should match with the input size of the TensorFlow Lite audio model that will consume the created TensorAudio object. If the WAV file contains more samples than `sample_coubt`, only the samples at the beginning of the WAV file will be loaded.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
kinaryml: @khanhlvg I've ported the Java implementation to Python. It should look something like this, right?

```python
def load_from_array(self,
                      src: np.ndarray,
                      offset: int = None,
                      size: int = None) -> None:
    """Loads the entire float NumPy array or a slice to the ring buffer."""
    if len(src) != len(self._buffer):
      raise ValueError(
        f"Input audio contains an invalid number of samples. "
        f"Expect {len(self._buffer)}.")
    elif src.shape[1] != self._format.channels:
      raise ValueError(
        f"Input audio contains an invalid number of channels. "
        f"Expect {self._format.channels}.")

    if offset is None and size is None:
      # Loads the entire float array to the ring buffer. If the float array is
      # longer than ring buffer's capacity, samples with lower indices in the
      # array will be ignored.
      self._buffer = src
    else:
      # Loads a slice of the float array to the ring buffer. If the float array
      # is longer than ring buffer's capacity, samples with lower indices in
      # the array will be ignored.
      buffer_size = len(self._buffer)
      next_index = self._next_index

      if size > buffer_size:
        offset = size - buffer_size
        size = buffer_size

      if next_index + size < buffer_size:
        # No need to wrap `next_index`.
        self._buffer[next_index:next_index + size] = src[offset:offset + size]
      else:
        # Need to wrap `next_index`, perform copy in two chunks.
        first_chunk_size = buffer_size - next_index
        self._buffer[next_index:buffer_size] = \
          src[offset:offset + first_chunk_size]
        self._buffer[0:size - first_chunk_size] = \
          src[offset + first_chunk_size:offset + size]

      self._next_index = (next_index + size) % buffer_size
```

khanhlvg: @kinaryml 
* The TensorAudio buffer size can be different to the src array, hence you shouldn't raise a ValueError when they're different.
* You should check if offset + size >= len(src) and raise a ValueError if it is.
* The size of the TensorAudio buffer shouldn't change after loading an array. Therefore, you can't do `self._buffer = src` but you have to check:
   * size > len(self._buffer): loading only the buffer size worth of samples from the end of the src array (`new_offset = offset + size - len(self._buffer)` and `new_size = len(self._buffer)`).
   * size < len(self._buffer): load the array into the end of the buffer. 
* Regarding the implementation of the buffer, you can use np.roll to make the code easier to read. There's no need to implement the RingBuffer exactly like the Java (storing the buffer begin index as next_index) because Java doesn't have convenient method like np.roll.

kinaryml: @khanhlvg Thanks for the feedback. Yeah the `offset + size > buffer_size` was a typo. Fixed it now. I'll take a look at the rest.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,158 @@
...
+    if record.sampling_rate != self._format.sample_rate:
+      raise ValueError(
+        f"The audio record's sampling rate doesn't match. "
+        f"Expects {self._format.sample_rate}Hz.")
+
+    # Load audio data from the AudioRecord instance.
+    data = record.read(self._buffer_size)
+    self.load_from_array(data.astype(np.float32))
+
+  def load_from_array(self,
```
kinaryml, line 102: @khanhlvg I've made some changes here now.

tensorflow_lite_support/python/test/task/audio/core/tensor_audio_test.py
```diff
@@ -0,0 +1,222 @@
...
+    audio_buffer = self.test_tensor_audio.buffer
+    audio_format = self.test_tensor_audio.format
+
+    self.assertEqual(audio_format.channels, _CHANNELS)
+    self.assertEqual(audio_format.sample_rate, _SAMPLE_RATE)
+    self.assertEqual(self.test_tensor_audio.buffer_size, _BUFFER_SIZE)
+    self.assertIsInstance(audio_buffer, np.ndarray)
+    testing.assert_almost_equal(audio_buffer, array)
+
+  def test_load_from_array_succeeds_with_larger_input_size(self):
```
kinaryml, line 74: @khanhlvg Added tests for loading arbitrary-sized arrays.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -64,7 +68,7 @@ def create_from_wav_file(cls,
     # `from pybind11_abseil import status`
     # see https://github.com/pybind/pybind11_abseil#abslstatusor.
     audio = _LoadAudioBufferFromFile(
-      file_name, buffer_size, np.zeros([buffer_size]))
+      file_name, sample_count, np.zeros([sample_count]))
```
khanhlvg, line 71: Can you test what happens when loading a WAV file with 2 channels or more? It looks like the array is resized in the C++ so I wonder if we need to reshape the numpy array after loading the WAV file to account for the channel dimension.

https://github.com/tensorflow/tflite-support/blob/e109aa0094e6fd5bb1e30c61ec9b19194c452ac3/tensorflow_lite_support/cc/task/audio/utils/wav_io.cc#L208

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -95,26 +99,48 @@ def load_from_audio_record(self, record: audio_record.AudioRecord) -> None:
...
+      src: A NumPy source array contains the input audio.
+      offset: An optional offset for loading a slice of the `src` array to the
+        buffer.
+      size: An optional size parameter denoting the number of samples to load
+        from the `src` array.
     Raises:
-      ValueError if the input audio is too large or if it contains an invalid
-      number of channels.
+      ValueError: Raised if the input array has an incorrect shape or if
+      offset + size exceeds the length of the `src` array.
```
khanhlvg, line 116: nit: put all parameters in a quote for API doc rendering later (e.g. \`offset\`)

kinaryml: @khanhlvg Will do.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -95,26 +99,48 @@ def load_from_audio_record(self, record: audio_record.AudioRecord) -> None:
...
+    if offset + size > len(src):
+      raise ValueError(
+        f"Index out of range. offset {offset} + size {size} should be <= "
+        f"src's length: {len(src)}")
+
+    if len(src) > len(self._buffer):
+      # Copy values from the source array to the internal buffer.
+      new_offset = offset + size - len(self._buffer)
+      new_size = len(self._buffer)
+      self._buffer = src[new_offset:new_offset + new_size]
```
khanhlvg, line 137: You should copy the values of the src array instead of holding a pointer to it.
`self._buffer = src[new_offset:new_offset + new_size].copy()`

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -95,26 +99,48 @@ def load_from_audio_record(self, record: audio_record.AudioRecord) -> None:
...
+    else:
+      # Shift the internal buffer backward and add the incoming data to the end
+      # of the buffer.
+      shift = len(src)
+      self._buffer = np.roll(self._buffer, -shift, axis=0)
+      self._buffer[-shift:, :] = src
```
khanhlvg, line 143: ditto: self._buffer[-shift:, :] = src.copy()

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -95,26 +99,48 @@ def load_from_audio_record(self, record: audio_record.AudioRecord) -> None:
...
+      size = len(src)
+    if offset is None:
+      offset = 0
+
+    if offset + size > len(src):
+      raise ValueError(
+        f"Index out of range. offset {offset} + size {size} should be <= "
+        f"src's length: {len(src)}")
+
+    if len(src) > len(self._buffer):
```
khanhlvg, line 133: Shouldn't this be `if len(src) >= len(self._buffer)`?

kinaryml: @khanhlvg Uh no because the opposite case is `<= newData.length` [here](https://github.com/tensorflow/tflite-support/blob/e109aa0094e6fd5bb1e30c61ec9b19194c452ac3/tensorflow_lite_support/java/src/java/org/tensorflow/lite/support/audio/TensorAudio.java#L304).

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
kinaryml: @khanhlvg Fixed it.

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
kinaryml: @khanhlvg I tried it out on a 16-bit stereo WAV file like [this](https://freewavesamples.com/files/Ensoniq-ZR-76-01-Dope-77.wav) one here just now and it works fine.

kinaryml: https://freewavesamples.com/ensoniq-zr-76-01-dope-77-bpm

kinaryml: @khanhlvg The returned shape for `tensor.buffer` NumPy array turns out to be this.
image

khanhlvg: Interesting. What happens if you pass in a np.zeros([1]) to the _LoadAudioBufferFromFile function? If it still works, then let use np.zeros([1]) and add a comment that the numpy array will be resized to make it clear for other maintainers of the library.

Then as a sanity test, could you play back the numpy array (e.g. using sounddevice.play) and check if the audio sounds correct?

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
khanhlvg: LGTM. Thanks!

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
kinaryml: @khanhlvg It doesn't seem to work with `np.zeros([1])`. The test crashes. On the other hand, I'll check if the audio comes out fine when loaded.

kinaryml: @khanhlvg I just tested how it sounds using the following snippet and I can hear the same beat on my end.
```python
import sounddevice as sd
sd.play(tensor.buffer)
time.sleep(10)
```

lu-wang-g (approved):
Looks great! Thanks for creating the PR, Kinar! And thanks for the detailed review, Khanh and Yuqi!

tensorflow_lite_support/python/task/audio/core/tensor_audio.py
```diff
@@ -0,0 +1,158 @@
...
+    if len(src) >= len(self._buffer):
+      # Copy values from the source array to the internal buffer.
+      new_offset = offset + size - len(self._buffer)
+      new_size = len(self._buffer)
+      self._buffer = src[new_offset:new_offset + new_size].copy()
+    else:
+      # Shift the internal buffer backward and add the incoming data to the end
+      # of the buffer.
+      shift = len(src)
+      self._buffer = np.roll(self._buffer, -shift, axis=0)
```
khanhlvg, line 142: There's a bug that you don't consider the offset here. I fixed it and also updated the test so that it can identify this bug if it's reintroduced. Please pull the latest code from master before continue the development.

PR #784, opened 2022-04-12
Migrating Python Task API tests to use tf.test along with some minor fixes

kinaryml:
- Updates copyright to the right year in some files of the vision tasks. 
- Migrate tests in Task Library to use `tf.test` instead of `unittest`
- Updated docstrings for Audio Task APIs' `required_audio_format` property

### DIFF
--- a/tensorflow_lite_support/python/test/base_test.py
+++ b/tensorflow_lite_support/python/test/base_test.py
@@ -13,12 +13,12 @@
 # limitations under the License.
 """Base TestCase for the unit tests."""
 
-import unittest
+import tensorflow as tf
 
 __unittest = True  # Allows shorter stack trace for .assertDeepAlmostEqual  pylint: disable=invalid-name
 
 
-class BaseTestCase(unittest.TestCase):
+class BaseTestCase(tf.test.TestCase):
   """Base test case."""
 
   def assertDeepAlmostEqual(self, expected, actual, **kwargs):
--- a/tensorflow_lite_support/python/task/audio/audio_classifier.py
+++ b/tensorflow_lite_support/python/task/audio/audio_classifier.py
@@ -132,5 +132,9 @@ class AudioClassifier(object):
 
   @property
   def required_audio_format(self) -> _CppAudioFormat:
-    """Gets the required audio format for the model."""
+    """Gets the required audio format for the model.
+
+    Raises:
+      RuntimeError: If failed to get the required audio format.
+    """
     return self._classifier.get_required_audio_format()
--- a/tensorflow_lite_support/python/task/audio/audio_embedder.py
+++ b/tensorflow_lite_support/python/task/audio/audio_embedder.py
@@ -151,5 +151,9 @@ class AudioEmbedder(object):
 
   @property
   def required_audio_format(self) -> _CppAudioFormat:
-    """Gets the required audio format for the model."""
+    """Gets the required audio format for the model.
+
+    Raises:
+      RuntimeError: If failed to get the required audio format.
+    """
     return self._embedder.get_required_audio_format()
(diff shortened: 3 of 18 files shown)


kinaryml:
@khanhlvg I've updated several tests and also moved `TensorImageTest` to the `python/task/vision/core` namespace to align with the other tasks. Let me know if you need any more changes.

khanhlvg (approved):

tensorflow_lite_support/python/test/base_test.py
```diff
@@ -13,12 +13,12 @@
...
+import tensorflow as tf
 
 __unittest = True  # Allows shorter stack trace for .assertDeepAlmostEqual  pylint: disable=invalid-name
 
 
-class BaseTestCase(unittest.TestCase):
+class BaseTestCase(tf.test.TestCase):
   """Base test case."""
 
   def assertDeepAlmostEqual(self, expected, actual, **kwargs):
```
khanhlvg, line 24: @kinaryml 
The reason to use tf.test, is to be able to use [https://www.tensorflow.org/api_docs/python/tf/test/TestCase#assertProtoEquals](https://www.google.com/url?sa=D&q=https%3A%2F%2Fwww.tensorflow.org%2Fapi_docs%2Fpython%2Ftf%2Ftest%2FTestCase%23assertProtoEquals) instead of our own implementation. Could you investigate if you can migrate away from using assertDeepAlmostEqual?

cc: @lu-wang-g

khanhlvg: @kinaryml I filed this issue for tracking as it'll be a large change. Please start investigating but there's no need to work on it in this PR.
https://github.com/tensorflow/tflite-support/issues/789

kinaryml: @khanhlvg Alright, I'll investigate it and let you know if I find anything.

kinaryml: @khanhlvg I believe [`assertProtoEquals`](https://github.com/tensorflow/tensorflow/blob/3f878cff5b698b82eea85db2b60d65a2e320850e/tensorflow/python/framework/test_util.py#L2465-L2488) cannot a floating-point comparison for numeric values with a delta like how [assertAllClose](https://github.com/tensorflow/tensorflow/blob/3f878cff5b698b82eea85db2b60d65a2e320850e/tensorflow/python/framework/test_util.py#L2919) can. We'd probably need a proto variant of the same in the TensorFlow library. 

I encountered this error if I try to compare a proto created using test data and the result proto.

`bash
AssertionError: Can't compare protos of type <class 'ClassificationResult'> and <class 'tensorflow_lite_support.cc.task.processor.proto.classifications_pb2.ClassificationResult'>.
`

khanhlvg: The error sounds like it doesn't allow comparing protos of different type. What happen if you convert ClassificationResult to tensorflow_lite_support.cc.task.processor.proto.classifications_pb2.ClassificationResult and compare them again?

kinaryml: @khanhlvg I tried the following to compare them and the tests pass.
```python
...

# Builds test data.
    expected_result = _build_test_data(expected_categories)

typed_image_result = classifications_pb2.ClassificationResult()
typed_image_result.ParseFromString(image_result.SerializeToString())
self.assertProtoEquals(typed_image_result, expected_result)
```
But this still doesn't allow us to specify `_ACCCEPTABLE_ERROR_RANGE` to account for floating-point errors.

kinaryml: For reference, sharing the `NormalizeNumberFields` function used by `assertProtoEquals`

https://github.com/tensorflow/tensorflow/blob/cec1bef0dbc1d42dbf082c86a86ae5de8625859c/tensorflow/python/util/protobuf/compare.py#L156

khanhlvg:
Merged.

PR #785, opened 2022-04-13
Remaining tests for the Python Image Segmenter Task Library

kinaryml:
- Fixed a `SegmentationOptions` initialization issue in `ImageSegmenterOptions`
- Added tests for category mask and confidence mask along with some additional tests

### DIFF
--- a/tensorflow_lite_support/python/test/task/vision/image_segmenter_test.py
+++ b/tensorflow_lite_support/python/test/task/vision/image_segmenter_test.py
@@ -280,7 +280,7 @@ class ImageSegmenterTest(parameterized.TestCase, tf.test.TestCase):
     segmentation = segmenter.segment(image).segmentation[0]
     category_mask = np.array(bytearray(segmentation.category_mask))
 
-    # Run segmentation on the model in CATEGORY_MASK mode.
+    # Run segmentation on the model in CONFIDENCE_MASK mode.
     segmenter = _create_segmenter_from_options(
       base_options, output_type=_OutputType.CONFIDENCE_MASK)
 
--- a/tensorflow_lite_support/python/test/task/vision/image_segmenter_test.py
+++ b/tensorflow_lite_support/python/test/task/vision/image_segmenter_test.py
@@ -15,6 +15,7 @@
 
 import enum
 
+import numpy as np
 from absl.testing import parameterized
 import tensorflow as tf
 from tensorflow_lite_support.python.task.core.proto import base_options_pb2
@@ -24,6 +25,7 @@ from tensorflow_lite_support.python.task.vision.core import tensor_image
 from tensorflow_lite_support.python.test import test_util
 
 _BaseOptions = base_options_pb2.BaseOptions
+_OutputType = segmentation_options_pb2.SegmentationOptions.OutputType
 _ImageSegmenter = image_segmenter.ImageSegmenter
 _ImageSegmenterOptions = image_segmenter.ImageSegmenterOptions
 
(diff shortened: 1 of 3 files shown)


tensorflow_lite_support/python/test/task/vision/image_segmenter_test.py
```diff
@@ -196,6 +231,80 @@ def test_segment_model(self, model_file_type, expected_colored_labels):
...
+    inconsistent_pixels = 0
+
+    for index in range(num_pixels):
+      inconsistent_pixels += (
+          result_pixels[index] * _MASK_MAGNIFICATION_FACTOR !=
+          ground_truth_pixels[index])
+
+    self.assertLessEqual(inconsistent_pixels / num_pixels,
+                         _MATCH_PIXELS_THRESHOLD,
+                         "Segmentation mask value must be the same size as ground truth.")
```
khanhlvg, line 265: The error message is incorrect here.

Besides, you should also assert if the segmentation mask size is the same as the ground truth mask size before the for loop.

khanhlvg (changes requested):
I've left some comments. PTAL. Thanks!

tensorflow_lite_support/python/test/task/vision/image_segmenter_test.py
```diff
@@ -196,6 +231,85 @@ def test_segment_model(self, model_file_type, expected_colored_labels):
...
+    result_pixels = np.array(bytearray(segmentation.category_mask))
+
+    # Loads ground truth segmentation file.
+    gt_segmentation = tensor_image.TensorImage.create_from_file(
+      self.test_seg_path)
+    gt_segmentation_array = gt_segmentation.buffer
+    gt_segmentation_shape = gt_segmentation_array.shape
+    num_pixels = gt_segmentation_shape[0] * gt_segmentation_shape[1]
+    ground_truth_pixels = gt_segmentation_array.flatten()
+
```
kinaryml, line 255: @khanhlvg I've added a check here before running the loop and also updated the error message to a more appropriate one. Let me know if you'd like to revise it.

khanhlvg: LGTM. Thanks!

tensorflow_lite_support/python/test/task/vision/image_segmenter_test.py
```diff
@@ -196,6 +231,85 @@ def test_segment_model(self, model_file_type, expected_colored_labels):
...
+                     'categories.')
+
+    # Gather the confidence masks in a single array `confidence_mask_array`.
+    confidence_mask_array = np.array([
+      confidence_masks[index].value for index in range(len(confidence_masks))])
+
+    # Compute the category mask from the created confidence mask.
+    calculated_category_mask = np.argmax(confidence_mask_array, axis=0)
+    self.assertListEqual(calculated_category_mask.tolist(),
+                         category_mask.tolist())
```
khanhlvg, line 311: Please add a message here.

tensorflow_lite_support/python/test/task/vision/image_segmenter_test.py
```diff
@@ -196,6 +231,85 @@ def test_segment_model(self, model_file_type, expected_colored_labels):
...
+    # Performs image segmentation on the input again.
+    segmentation = segmenter.segment(image).segmentation[0]
+
+    # Gets the list of confidence masks and colored_labels.
+    confidence_masks = segmentation.confidence_masks.confidence_mask
+    colored_labels = segmentation.colored_labels
+
+    # Check if confidence mask shape is correct.
+    self.assertEqual(len(confidence_masks), len(colored_labels),
+                     'Number of confidence masks must match with number of '
```
khanhlvg, line 301: Please don't mix single quote and double quote in the same file. Let's unify to double quote then.

khanhlvg (approved):

PR #787, opened 2022-04-13
iOS Object Detection Code Refactor

priankakariatyml:
1. Code refactor for object detection.
2. Method signature change in image classifier and corresponding change in tests


tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.m
```diff
@@ -39,7 +40,7 @@ - (instancetype)init {
   return self;
 }
 
-- (nullable instancetype)initWithModelPath:(nonnull NSString *)modelPath {
+- (instancetype)initWithModelPath:(NSString *)modelPath {
```
priankakariatyml, line 43: @khanhlvg I have removed the nullability of the return type here since the model path argument is non nullable by default and hence under no circumstances the created options can be null. But this holds true in swift since swift compiler generates an error if a null is passed. In objC only a warning is shown by the compiler. Theoretically you can still pass a null in ObjC and that ends up making the options non usable. So in such a case should we do a null check and return null for options or not handle this case?

tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.m
```diff
@@ -62,8 +63,16 @@ - (instancetype)initWithObjectDetector:(TfLiteObjectDetector *)objectDetector {
   return self;
 }
 
-+ (nullable instancetype)objectDetectorWithOptions:(nonnull TFLObjectDetectorOptions *)options
++ (nullable instancetype)objectDetectorWithOptions:(TFLObjectDetectorOptions *)options
                                              error:(NSError **)error {
+
+  if (!options) {
```
priankakariatyml, line 69: I have added a null check for options here. Please see my comments [here](https://github.com/tensorflow/tflite-support/pull/787/files#r849323481) and LMK if this is needed win practice.

tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.m
```diff
@@ -88,6 +97,14 @@ + (nullable instancetype)objectDetectorWithOptions:(nonnull TFLObjectDetectorOpt
 
 - (nullable TFLDetectionResult *)detectWithGMLImage:(GMLImage *)image
                                               error:(NSError *_Nullable *)error {
+
+  if (!image) {
```
priankakariatyml, line 101: @khanhlvg Thoughts on this as well. Based on my comment [here](https://github.com/tensorflow/tflite-support/pull/787/files#r849323481)  and [here](https://github.com/tensorflow/tflite-support/pull/787/files#r849324803).

tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.m
priankakariatyml: Also should we make options of all tasks subclasses of TFLBaseOptions? TFLBaseOptions is common for all tasks. In this way they'll all have a consistent initWithModelPath. So modifications can be done in one place.

khanhlvg: > Theoretically you can still pass a null in ObjC and that ends up making the options non usable. So in such a case should we do a null check and return null for options or not handle this case?

You can return a null object as if a null modelPath is passed to the initializer.

> Also should we make options of all tasks subclasses of TFLBaseOptions? TFLBaseOptions is common for all tasks. In this way they'll all have a consistent initWithModelPath. So modifications can be done in one place.

No. Let's keep the structure as it: ObjectDetectorOptions contains 2 properties: a TFLBaseOptions and a TFLClassificationOptions.. It aligns with the C++ and Python APIs.

tensorflow_lite_support/ios/task/processor/sources/TFLDetectionResult.h
```diff
@@ -26,18 +26,39 @@ NS_ASSUME_NONNULL_BEGIN
  * The index of the image classifier head these classes refer to. This is useful for multi-head
  * models.
  */
-@property(nonatomic, assign) CGRect boundingBox;
+@property(nonatomic, assign, readonly) CGRect boundingBox;
 
 /** The array of predicted classes, usually sorted by descending scores (e.g.from high to low
  * probability). */
-@property(nonatomic, copy) NSArray<TFLCategory *> *categories;
+@property(nonatomic, copy, readonly) NSArray<TFLCategory *> *categories;
```
khanhlvg, line 33: As the property is readonly, you should have "copy" there.

tensorflow_lite_support/ios/task/vision/sources/TFLImageClassifier.h
```diff
@@ -50,7 +50,7 @@ NS_ASSUME_NONNULL_BEGIN
  * @return An instance of TFLImageClassifierOptions set to the specified
  * modelPath.
  */
-- (nullable instancetype)initWithModelPath:(NSString *)modelPath;
+- (instancetype)initWithModelPath:(NSString *)modelPath;
```
khanhlvg, line 53: @lu-wang-g @mbrenon After implementing the Python API, I realized that this default initializer should be implemented in the ImageClassifer rather than the ImageClassifierOptions. So we can do a one-liner:

```objc
TFLImageClassifier *classifier = [TFLImageClassifier imageClassifierWithModelPath:@"model.tflite" error:nil];
```

rather than 2 liner:
```objc
TFLImageClassifierOptions options = [[TFLImageClassifierOptions alloc] initWithModelPath:@"model.tflite"];
TFLImageClassifier *classifier = [TFLImageClassifier imageClassifierWithOptions:@"model.tflite" error:nil];
```

This will align better with the Python and Java API. WDYT?

tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.m
khanhlvg: Yes that makes sense.

tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.m
khanhlvg: Yes, makes sense.

tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.h
```diff
@@ -49,7 +49,7 @@ NS_ASSUME_NONNULL_BEGIN
  * @return An instance of TFLObjectDetectorOptions set to the specified
  * modelPath.
  */
-- (nullable instancetype)initWithModelPath:(nonnull NSString *)modelPath;
+- (instancetype)initWithModelPath:(NSString *)modelPath;
```
khanhlvg, line 52: We should still mark modelPath as nonnull here, and the `(nullable instancetype)objectDetectorWithOptions:(nonnull TFLObjectDetectorOptions *)options` below. ObjC doesn't enforce the null check at compile time and that's fine. If an error is created and the objectDetectorWithOptions:error: returns nil, the user can examine the error there.

priankakariatyml: modelPath is automatically considered nonnull because of NS_ASSUME_NONNULL_BEGIN in the headers. Only nullable args need to be marked explicitly as nullable.  So compiler will still throw an error if this is null in swift.

Also my doubt is whether to add the nullable to return type here, since modelPath can be null only in ObjC even if it is marked  as nonnull. 
In that case there are 2 options:
 1. Initialise the baseOptions.modelFile.modelPath to null and let TFLObjectDetectorOptions be initialised. In this case return type will be nonnull. 
 2. Check for nullability of modelPath and return a null TFLObjectDetectorOptions. In this case return type will be nullable.  We can add an error to the initialiser as well here.

tensorflow_lite_support/ios/task/vision/sources/TFLImageClassifier.h
priankakariatyml: Right now TFLImageClassifierOptions is not a property or instance variable of TFLImageClassifier. So with the suggested method signature we will have to add the image classifier options as a readonly property, since our C Image Classifier initialisation will be completed once the ObjC Image Classifier init completes and we can't allow changes after that. 
So if this signature is to be used we will need 2 init methods:
1. TFLImageClassifier *classifier = [TFLImageClassifier imageClassifierWithModelPath:@"model.tflite" error:nil];
If you use this initialiser you won't be able to modify any other options. It will be like using default options with a modelPath. 
2.  TFLImageClassifier *classifier = [TFLImageClassifier imageClassifierWithOptions:options error:nil];
This will let us configure options acc convenience and init. From thereon, options will be only readable.

tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.h
priankakariatyml: Disregard my comments on the null check. I saw your response.

tensorflow_lite_support/ios/task/processor/sources/TFLDetectionResult.m
```diff
@@ -16,13 +16,26 @@
...
 @implementation TFLDetection
 
-@synthesize boundingBox;
-@synthesize categories;
+- (instancetype)initWithBoundingBox:(CGRect)boundingBox
+                         categories:(NSArray<TFLCategory *> *)categories {
+  self = [super init];
+  if (self) {
+    _boundingBox = boundingBox;
+    _categories = [categories copy];
```
priankakariatyml, line 24: @khanhlvg The copy here is not a deep copy. It only prevents any changes if the passed in categories arg is an NSMutableArray. The properties of TFLCategory objects are a shallow copy. But since our TFLCategory object properties are readonly, they can't really be modified. So should we loop through and  do a copy of each element or is that unnecessary?

tensorflow_lite_support/ios/task/vision/sources/TFLObjectDetector.h
khanhlvg: never mind the nonnull comment. I forgot that notnull is the defaut.

tensorflow_lite_support/ios/task/processor/sources/TFLDetectionResult.m
khanhlvg: No that's fine. Excessive copy will increase the latency.

tensorflow_lite_support/ios/task/processor/sources/TFLDetectionResult.h
khanhlvg, line 33: since you're dealing with the synthesized variable directly rather than using the property in the implementation, it's fine to remove `copy` here as you're not using it.

priankakariatyml: Removed

tensorflow_lite_support/ios/task/vision/sources/TFLImageClassifier.m
```diff
@@ -79,8 +87,11 @@ + (nullable instancetype)imageClassifierWithOptions:(TFLImageClassifierOptions *
   [options.classificationOptions
       deleteCStringArraysOfClassificationOptions:&(cOptions.classification_options)];
 
-  if (!imageClassifier || ![TFLCommonUtils checkCError:createClassifierError toError:error]) {
```
priankakariatyml, line 82: @khanhlvg  I guess the implementation of TFLCommonUtils was modified slightly from my version after the error refactor PR was merged. When I checked with my latest tests, this condition check was causing the error to not be populated as LHS evaluates to true and hence or condition's RHS was not being evaluated. I have modified the condition checks to cover all bases for Image Classifier and Object Detector in this PR. Image segmentation has been modified in its respective code refactor PR.

tensorflow_lite_support/ios/task/processor/sources/TFLClassificationResult.m
```diff
@@ -21,7 +21,7 @@ - (instancetype)initWithHeadIndex:(NSInteger)headIndex
   self = [super init];
   if (self) {
     _headIndex = headIndex;
-    _categories = categories;
+    _categories = [categories copy];
```
khanhlvg, line 24: After the benchmark I did when designing the ImageSegmenter output, I learnt that memory copy here can become quite expensive. So to avoid excessive copy, let's keep the strong reference of the value as you're doing now rather than copy them.

tensorflow_lite_support/ios/task/vision/sources/TFLImageClassifier.h
khanhlvg: @priankakariatyml Yes that's what I expect it to be. Let's see what Lu and Maxime think about this new API. If they agrees, let's migrate the API so that we'll reduce the amount of code to write for users, and also have consistency between languages.

tensorflow_lite_support/ios/task/vision/sources/TFLImageClassifier.m
khanhlvg: Yes it was pointed out in the internal review so I changed it. Thanks for catching the issue and fixed it.

tensorflow_lite_support/ios/task/processor/sources/TFLClassificationResult.m
priankakariatyml: Fixed this. reverted for all string and array copies

khanhlvg (approved):

PR #788, opened 2022-04-13
iOS Image Segmentation Code Refactor

priankakariatyml:
1. Code refactor for Image Segmenter.
2. Added NS_SWIFT_NAME for Image Segmenter.


tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
```diff
@@ -13,86 +13,99 @@
...
+}
 
 @end
 
 @implementation TFLSegmentationResult
-@synthesize segmentations;
+- (instancetype)initWithSegmentations:(NSArray<TFLSegmentation *> *)segmentations {
+  self = [super init];
+  if (self) {
+    _segmentations = [segmentations copy];
```
khanhlvg, line 106: Let's avoid excessive copy here. The only place where we have to create a copy is when converting from C object to ObjC object (e.g. (float *)mask to TFLConfidenceMask) because we need to manage the ownership of the object. Otherwise, just keeping a strong reference should be enough because the data type used is immutable.

priankakariatyml: Fixed. Reverted copying for all strings and arrays.

tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
```diff
@@ -13,86 +13,99 @@
...
+      _mask = [TFLCommonUtils mallocWithSize:width * height * sizeof(UInt8) error:nil];
+      if (_mask) {
+        memcpy(_mask, mask, width * height * sizeof(UInt8));
+      }
     }
   }
   return self;
 }
 
-- (id)copyWithZone:(NSZone *)zone {
```
khanhlvg, line 36: Please keep the copy method as some users may want to use it.

tensorflow_lite_support/ios/task/processor/sources/TFLSegmentationResult.m
```diff
@@ -13,86 +13,99 @@
...
+      _mask = [TFLCommonUtils mallocWithSize:width * height * sizeof(float) error:nil];
+      if (_mask) {
+        memcpy(_mask, mask, width * height * sizeof(float));
+      }
     }
   }
   return self;
 }
 
-- (id)copyWithZone:(NSZone *)zone {
```
khanhlvg, line 67: Please keep the copy method as some users may want to use it.

PR #793, opened 2022-04-16
Task Library: Migrate away from `assertDeepAlmostEqual` 

kinaryml:
- Removed file base_test.py and the `assertDeepAlmostEqual` test utility function used by any tests.
- Updated some tests to use protobuf text proto format instead of JSON for the expected result
- Minor fixes to naming


khanhlvg (approved):

khanhlvg:
Merged.

PR #798, opened 2022-04-22
Task Library: Python TextSearcher API Implementation

kinaryml:
- Added TextSearcher implementation.
- Includes a sanity test to check if the API is functioning properly.


tensorflow_lite_support/python/task/core/proto/BUILD
```diff
@@ -25,3 +25,11 @@ support_py_proto_library(
     api_version = 2,
     proto_deps = [":base_options_proto"],
 )
+
+py_library(
+    name = "external_file_pb2",
+    srcs = ["external_file_pb2.py"],
+    deps = [
+        "//tensorflow_lite_support/cc/task/core/proto:external_file_py_pb2",
```
khanhlvg, line 33: @mbrenon @lu-wang-g @ziyeqinghan Currently Python's [BaseOptions](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/python/task/core/proto/base_options.proto) doesn't use ExternalFile proto to define a model file, but SearchOptions does, creating inconsistency between the two options. Should we achieve consistency by either of these?
* Redefine Python's base_options.proto to use ExternalFile.
* Redefine search_options proto in Python layer to match with the Python's base_options style?

kinaryml: IMHO, we can avoid redefining [`BaseOptions`](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/python/task/core/proto/base_options.proto) in Python and reuse the BaseOptions and ExternalFile core protos from C++. The API should look something like this if we were to makethe.

```python
model_file = _ExternalFile(file_name=model_file_path)
index_file = _ExternalFile(file_name=index_file_path)
options = TextSearcherOptions(
  base_options=_BaseOptions(model_file=model_file),
  embedding_options=_EmbeddingOptions(l2_normalize=True),
  search_options=_SearchOptions(index_file=index_file))
...
```

khanhlvg: The C++ BaseOptions contains many flags that aren't relevant to Python, such as the Android NNAPI delegate or iOS CoreML delegate so we don't want to expose them directly. So that's why we redefined a Python specific BaseOptions.

kinaryml: @khanhlvg Then I suppose we should stick to the current style used by `BaseOptions` proto in Python and redefine `SearchOptions` in a similar fashion. I can make the changes to the API and it should look something like this.

```python
options = TextSearcherOptions(
  base_options=_BaseOptions(file_name=model_file_path),
  embedding_options=_EmbeddingOptions(l2_normalize=True),
  search_options=_SearchOptions(file_name=index_file_path))
...
```
WDYT?

kinaryml: > * Redefine Python's base_options.proto to use ExternalFile.

On a side note, this should make the API more readable since users can make sense of passing `model_file = _ExternalFile(file_name=model_file_path)` to `BaseOptions`'s `model_file` parameter rather than directly passing the file path using `_BaseOptions(file_name=model_file_path)`

kinaryml: Another option I can think of to make the existing implementation of `BaseOptions` proto in Python more readable would be to:
- Rename `file_name` and `file_content` in [`BaseOptions`](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/python/task/core/proto/base_options.proto) to `model_file_name` and `model_file_content` respectively.
- Define `index_file_name` and `index_file_content` similarly in Python's `SearchOptions` proto. 

This would avoid a lot of rework necessary in the Task APIs and I believe only the Task Core Utils [pybinds](https://github.com/tensorflow/tflite-support/blob/5a059e141f4a54fd802f14fe226af6decc255989/tensorflow_lite_support/python/task/core/pybinds/task_utils.cc#L31) would have to be updated to account for the change.

mbrenon: I would favor not exposing ExternalFile in the Python API, for consistency with what we did for BaseOptions. So, I suggest we create python/task/processor/proto/search_options.proto with:

```
message SearchOptions {
  optional string index_file_name = 1;
  optional string index_file_content = 2;
  optional int32 max_results = 3;
}
```

With the index file fields documented as they are in python/task/core/proto/base_options.proto.

IMHO modifying BaseOptions is not needed.

kinaryml: @mbrenon Alright, will do.

tensorflow_lite_support/python/task/processor/proto/BUILD
```diff
@@ -1,4 +1,4 @@
-# Placeholder for internal Python strict library compatibility macro.
```
khanhlvg, line 1: Please don't delete this line as it will get replaced automatically in internal build.

kinaryml: @khanhlvg Alright, will re-add it. However, I noticed it was removed in another commit sometime ago so decided to do the same here.

kinaryml: @khanhlvg Fixed it.

tensorflow_lite_support/python/task/processor/proto/search_options.proto
```diff
@@ -0,0 +1,35 @@
...
+package tflite.python.task.processor;
+
+option java_multiple_files = true;
+option java_package = "org.tensorflow.lite.task.processor.proto";
+
+// Options for Python search processor.
+// See C++ search options at:
+// https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/cc/task/processor/proto/search_options.proto
+// Next Id: 4
+message SearchOptions {
```
mbrenon, line 27: Can you add a comment similar to https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/python/task/core/proto/base_options.proto#L29-L38?

kinaryml: Updated it.

khanhlvg (approved):

PR #804, opened 2022-04-29
Updated C classification Result Struct

priankakariatyml:
Updated C Classification Result struct to add a new member head_name which is used in audio classifier APIs.

### DIFF
--- a/tensorflow_lite_support/c/task/processor/classification_result.h
+++ b/tensorflow_lite_support/c/task/processor/classification_result.h
@@ -29,6 +29,12 @@ typedef struct TfLiteClassifications {
   // useful for multi-head models.
   int head_index;
 
+  // The name of the classifier head, which is the corresponding tensor metadata
+  // name. See
+  // https://github.com/tensorflow/tflite-support/blob/710e323265bfb71fdbdd72b3516e00cff15c0326/tensorflow_lite_support/metadata/metadata_schema.fbs#L545
+  // This will always be NULL for vision APIs.
+  char* head_name;
+
   // Number of predicted classes which can be used to traverse the array of
   // predicted classes.
   int size;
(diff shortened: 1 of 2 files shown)


tensorflow_lite_support/c/task/processor/classification_result.h
```diff
@@ -29,6 +29,12 @@ typedef struct TfLiteClassifications {
   // useful for multi-head models.
   int head_index;
 
+  // The name of the classifier head, which is the corresponding tensor metadata
+  // name. See
+  // https://github.com/tensorflow/tflite-support/blob/710e323265bfb71fdbdd72b3516e00cff15c0326/tensorflow_lite_support/metadata/metadata_schema.fbs#L545
+  // This will always be NULL for vision APIs.
+  char* head_name;
```
priankakariatyml, line 36: @lu-wang-g @khanhlvg @mbrenon  In C, since the ClassificationResult struct is common for all classification tasks, I am adding head_name to the struct to account for audio_classifier APIs. C++ audio classifier uses this [proto](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/cc/task/core/proto/classifications.proto) as opposed to image classifier [counterpart](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/cc/task/vision/proto/classifications.proto) which has no head_name. 

Likewise this will have to be added for the iOS [TFLClassifications](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/ios/task/processor/sources/TFLClassificationResult.h#L22) too. But can this be done before release for iOS since nightly is already out? Otherwise we will have to create a new TFLAudioClassificationResult for the audio classifier to account for head_name.

khanhlvg (approved):

PR #808, opened 2022-05-05
Added Audio Tensor And Audio Record

priankakariatyml:


tensorflow_lite_support/ios/task/audio/core/sources/TFLRingBuffer.h
```diff
@@ -0,0 +1,74 @@
...
+ */
+- (BOOL)loadBuffer:(TFLFloatBuffer *)sourceBuffer
+            offset:(NSUInteger)offset
+              size:(NSUInteger)size
+             error:(NSError **)error;
+
+/**
+ * Returns a `TFLFloatBuffer` with a copy of all the ring buffer elements in order.
+ */
+- (TFLFloatBuffer *)floatBuffer NS_SWIFT_NAME(floatBuffer());
```
khanhlvg, line 48: Define this as a property rather than a method.

khanhlvg (changes requested):
This PR is quite large so please split it into smaller PRs of less than 400 lines each to make internal review easier.

Besides, please add unit tests for the classes other than TFLAudioRecord.

PR #814, opened 2022-05-12
Added C Audio Classifier Implementation and Basic Tests

priankakariatyml:
1. Added C audio classifier implementation files.
2. Added basic tests


khanhlvg (approved):

khanhlvg:
Some changes needed when adapting to internal codebase:
* Use `char *const` instead of `char *` to be able to accept string literals.
* Use `.cc` instead of `.c` for pure C code (this is unique to how Task Library is setup)
* Added missing dependencies.

PR #815, opened 2022-05-12
Added iOS Ring Buffer and Float Buffer

priankakariatyml:


tensorflow_lite_support/ios/task/audio/core/sources/TFLRingBuffer.m
```diff
@@ -0,0 +1,118 @@
...
+}
+
+- (nullable TFLFloatBuffer *)floatBufferWithOffset:(NSUInteger)offset size:(NSUInteger)size {
+  if (offset + size > _buffer.size) {
+    return nil;
+  }
+
+  // TODO: Investigate if allocating is needed everytime or a global alloc of self.floatBuffer and
+  // copying of elements each time and returning self.floatBuffer would suffice.
+  TFLFloatBuffer *bufferToReturn = [[TFLFloatBuffer alloc] initWithSize:size];
```
khanhlvg, line 86: It's necessary to alloc every time because the docstring says that this method returns a **copy** of the internal buffer in the correct order.

khanhlvg (approved):

khanhlvg:
Merged in https://github.com/tensorflow/tflite-support/commit/0020b50bff11205b5a246b85c5aad6dcd30e454e

PR #822, opened 2022-05-19
Added iOS Audio Record and Audio Format

priankakariatyml:


tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
```diff
@@ -0,0 +1,269 @@
...
+                                       error:(NSError *_Nullable *)error {
+  self = [self init];
+  if (self) {
+    if (audioFormat.channelCount > SUPPORTED_CHANNEL_COUNT) {
+      [TFLCommonUtils
+          createCustomError:error
+                 withDomain:TFLAudioRecordErrorDomain
+                       code:TFLAudioRecordErrorCodeInvalidArgumentError
+                description:@"The channel count provided does not match the supported "
+                            @"channel count. Only upto 2 audio channels are currently supported."];
```
khanhlvg, line 51: nit: As you defined SUPPORTED_CHANNEL_COUNT as a macro, it's better to use the macro value here instead of hard-code it in the error string.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
```diff
@@ -0,0 +1,269 @@
...
+                          @"input source. Please wait for the input."];
+
+    _globalError = waitError;
+    _audioFormat = audioFormat;
+    _audioEngine = [[AVAudioEngine alloc] init];
+    _bufferSize = bufferSize;
+
+    _ringBuffer = [[TFLRingBuffer alloc] initWithBufferSize:_bufferSize];
+    _conversionQueue =
+        dispatch_queue_create("com.tflAudio.AudioConversionQueue", NULL);  // Serial Queue
```
khanhlvg, line 79: Please use `org.tensorflow.lite.AudioConversionQueue` to correctly align with the TFLite namespace.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecordErrorCode.h
```diff
@@ -0,0 +1,55 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ ==============================================================================*/
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * @enum TFLAudioRecordErrorCode
+ * This enum specifies  error codes for TFLAudioRecord of the TensorFlow Lite Task Library.
```
khanhlvg, line 21: nit: remove double spaces between "specifies" and "error"

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecordErrorCode.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+typedef NS_ENUM(NSUInteger, TFLAudioRecordErrorCode) {
```
khanhlvg, line 23: Please move this to the TFLAudioRecord.h file because it's an integral part of the AudioRecord API.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
```diff
@@ -0,0 +1,269 @@
...
+                                 code:TFLAudioRecordErrorCodeInvalidArgumentError
+                          description:@"You may have to try with a different "
+                                      @"channel count or sample rate"];
+  } else if ((pcmBuffer.frameLength % self.audioFormat.channelCount) != 0) {
+    [TFLCommonUtils createCustomError:error
+                           withDomain:TFLAudioRecordErrorDomain
+                                 code:TFLAudioRecordErrorCodeInvalidArgumentError
+                          description:@"You have passed an unsupported number of channels."];
+  } else {
+    TFLFloatBuffer *floatBuffer = [[TFLFloatBuffer alloc] initWithData:pcmBuffer.floatChannelData[0]
```
khanhlvg, line 153: This looks like you're taking only the first channel from the pcmBuffer. In the case that AVAudioPCMBuffer has more than 1 channel, wouldn't the rest of the channel be filled with zeros?

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
```diff
@@ -0,0 +1,269 @@
...
+                                                                  size:pcmBuffer.frameLength];
+
+    if ([self->_ringBuffer loadBuffer:floatBuffer offset:0 size:floatBuffer.size error:error]) {
```
khanhlvg, line 156: You're doing memory copy twice here: from pcmBuffer.floatChannelData to floatBuffer then to _ringBuffer, which may cause added latency. This can be reduced to one copy by adding the `loadFloat:(float *)` API to the TFLRingBuffer.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
khanhlvg, line 153: pcmBuffer.floatChannelData only exists if the format is 32-bit float. Please add a check and create a custom error if it's not.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
```diff
@@ -0,0 +1,269 @@
...
+                                                                     toFormat:recordingFormat];
+
+  // Setting buffer size takes no effect on the input node. This class uses a ring buffer internally
+  // to ensure the requested buffer size.
+  [inputNode
+      installTapOnBus:0
+           bufferSize:(AVAudioFrameCount)self.bufferSize
+               format:format
+                block:^(AVAudioPCMBuffer *buffer, AVAudioTime *when) {
+                  dispatch_async(self->_conversionQueue, ^{
```
khanhlvg, line 183: Move this logic to a method so that you can write unit test for it without mocking the AVAudioNode.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
```diff
@@ -0,0 +1,269 @@
...
+    }
+  });
+
+  if (!bufferToReturn && error) {
+    *error = readError;
+  }
+  
+  return bufferToReturn;
+}
+
```
khanhlvg, line 268: Please write unit tests for this class. You can create a AVAudioPCMBuffer programmatically, feed it through the conversion pipeline and assert if `readAtOffset:withSize:error` returns a correct result. Please make sure to test both single-channel and double-channel input.

khanhlvg (changes requested):

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
priankakariatyml: We are requesting the audio converter to return channels in an interleaved format. In this case both floatChannelData[0] and floatChannelData[1] point to the same 1d array with both channels in an interleaved format. You can check the [doc](https://developer.apple.com/documentation/avfaudio/avaudiopcmbuffer/1386212-floatchanneldata). I also tested this out.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.h
```diff
@@ -0,0 +1,98 @@
...
+ * any given point of time. The buffer length should be a multiple of format.channelCount.
+ * @param error An optional error parameter populated if the initialization of `TFLAudioRecord` was
+ * not successful.
+ *
+ * @return An new instance of `TFLAudioRecord` with the given audio format and buffer size. `nil` if
+ * there is an error in initializing `TFLAudioRecord`.
+ */
+- (nullable instancetype)initWithAudioFormat:(TFLAudioFormat *)format
+                                  bufferSize:(NSUInteger)bufferSize
+                                       error:(NSError *_Nullable *)error;
```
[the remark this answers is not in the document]
priankakariatyml: Fixed. I had written this file before I made this change to the vision APIs and did not change it here. Hence the mismatch.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
khanhlvg: I see. Could you leave a comment in the doc to make it clearer? This behavior isn't obvious so we should leave a note for future maintainers of the library.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
priankakariatyml: I have changed the interface of TFLRingBuffer. Load can be done using `loadFloatData:(float *) dataSize:(NSUInteger)`. I have removed the load from `TFLFloatBuffer` as it won't be used by any of our APIs.  `TFLFloatBuffer` will only be used for returning the data using` readAtOffset`.

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
priankakariatyml: Fixed

tensorflow_lite_support/ios/task/audio/core/audio_record/sources/TFLAudioRecord.m
priankakariatyml: Fixed.

tensorflow_lite_support/ios/test/task/audio/core/audio_record/TFLAudioRecordTests.m
```diff
@@ -0,0 +1,296 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface TFLAudioRecordTests : XCTestCase
+@property(nonatomic) AVAudioFormat *audioEngineFormat;
+@end
+
+// This category of TFLAudioRecord is private to the current test file.
+// This is needed in order to expose the method to load the audio record buffer 
+// without calling: -[TFLAudioRecord startRecordingWithError:].
+// This is needed to avoid exposing this method which isn't useful to the consumers
```
priankakariatyml, line 37: @khanhlvg Swift tests cannot be added for loading and readAtOffset since there is no similar workaround to expose this private method to tests without exposing it in the framework too.

khanhlvg: That's fine. We only need to test the internal logic of TFLAudioRecord so having only ObjC tests is enough.

khanhlvg (approved):

PR #823, opened 2022-05-21
Task Library: Migrate Searcher task APIs to dataclasses

kinaryml:


tensorflow_lite_support/python/task/processor/proto/search_options_pb2.py
```diff
@@ -0,0 +1,85 @@
...
+  @doc_controls.do_not_generate_docs
+  def to_pb2(self) -> _SearchOptionsProto:
+    """Generates a protobuf object to pass to the C++ layer."""
+    if self.index_file_name is not None or self.index_file_content is not None:
+      return _SearchOptionsProto(
+        index_file=_ExternalFileProto(
+          file_name=self.index_file_name,
+          file_content=self.index_file_content),
+        max_results=self.max_results)
+
```
khanhlvg, line 60: raise a ValueError here if neither index_file_name or index_file_content is specified.

tensorflow_lite_support/python/task/text/pybinds/_pywrap_text_searcher.cc
```diff
@@ -57,19 +55,9 @@ PYBIND11_MODULE(_pywrap_text_searcher, m) {
...
-            }
-            if (search_options.has_index_file_name()) {
-              cpp_search_options->mutable_index_file()->set_file_name(
-                  search_options.index_file_name());
-            }
-            if (search_options.has_max_results()) {
-              cpp_search_options->set_max_results(search_options.max_results());
-            }
-
+            cpp_search_options->CopyFrom(search_options);
```
khanhlvg, line 58: nit: Can you use the search_options directly instead of having to copy it?

tensorflow_lite_support/python/test/task/vision/image_searcher_test.py
```diff
@@ -35,6 +35,28 @@
 _MOBILENET_EMBEDDER_MODEL = 'mobilenet_v3_small_100_224_embedder.tflite'
 _MOBILENET_SEARCHER_MODEL = 'mobilenet_v3_small_100_224_searcher.tflite'
 _MOBILENET_INDEX = 'searcher_index.ldb'
+_EXPECTED_MOBILENET_DEFAULT_OPTIONS_SEARCH_RESULT = """
```
khanhlvg, line 38: Please create a SearchResult object to define the expected results instead of using protobuf here.

tensorflow_lite_support/python/test/task/text/text_searcher_test.py
```diff
@@ -260,7 +260,7 @@ def test_search_with_default_options(self):
     text_search_result = searcher.search('The weather was excellent.')
 
     self.assertProtoEquals(_EXPECTED_REGEX_DEFAULT_OPTIONS_SEARCH_RESULT,
```
khanhlvg, line 262: Please convert the definition of expected results from proto text to defining them using SearchResult objects. Ideally you should compare the expected results with computed results using `self.assert()`. If that doesn't work because of floating point accuracy issue, you can fall back to using `.to_pb2()` and compare using `self.assertProtoEquals` like what you're doing now.

khanhlvg (changes requested):

tensorflow_lite_support/python/task/text/pybinds/_pywrap_text_searcher.cc
kinaryml: @khanhlvg I can change it, however, I believe in this [commit](https://github.com/tensorflow/tflite-support/commit/e109aa0094e6fd5bb1e30c61ec9b19194c452ac3), we decided to set the options by copying it from the source. WDYT?

tensorflow_lite_support/python/test/task/text/text_searcher_test.py
kinaryml: @khanhlvg I think we'll have to fall back to using `.to_pb2()` as it seems to compare at full precision instead of `1e-6` like how `assertProtoEquals` is able to do right now. Meanwhile, I've updated the expected result objects to use the SearchResult python objects instead of the protobuf text protos.

tensorflow_lite_support/python/test/task/vision/image_searcher_test.py
kinaryml: Fixed it

tensorflow_lite_support/python/task/processor/proto/search_options_pb2.py
kinaryml: @khanhlvg The C++ API seems to handle it already [here](https://github.com/tensorflow/tflite-support/blob/a1d17bf66ef1e148ac6bca2ec3bc81d9993d8421/tensorflow_lite_support/cc/task/core/external_file_handler.cc#L78) though and if both of them are unspecified, it just falls back to using just the `max_results` option [here](https://github.com/tensorflow/tflite-support/pull/823/files#diff-a5fd5f315b938ca84318b6a8b6bd2c2b9cf1e94861512af4699f18832ee6bf08R61).

khanhlvg: Then you should remove `if self.index_file_name is not None or self.index_file_content is not None:` because Python type check will complain that this function may return None, which doesn't match with the function's return type (_SearchOptionsProto)

tensorflow_lite_support/python/task/text/pybinds/_pywrap_text_searcher.cc
khanhlvg: You're right. Copying is needed from the memory management perspective.

tensorflow_lite_support/python/test/task/text/text_searcher_test.py
khanhlvg: Falling back to assertProtoEquals and .to_pb2() SGTM. Thanks for checking.

tensorflow_lite_support/python/task/processor/proto/search_result_pb2.py
```diff
@@ -13,7 +13,90 @@
...
+  """A single nearest neighbor.
+
+  Attributes:
+    metadata: User-defined metadata about the result. This could be a label, a
+      unique ID, a serialized proto of some sort, etc.
+    distance: The distance score indicating how confident the result is.
+      Lower is better.
+  """
+
+  metadata: bytearray
```
khanhlvg, line 37: Please convert this to string.

kinaryml: @khanhlvg Fixed it.

tensorflow_lite_support/python/task/processor/proto/search_options_pb2.py
kinaryml: @khanhlvg But I think a check to switch between the two modes (Embedder/Searcher) is necessary here. The Searcher model doesn't require an index file as illustrated [here](https://github.com/tensorflow/tflite-support/blob/a1d17bf66ef1e148ac6bca2ec3bc81d9993d8421/tensorflow_lite_support/python/test/task/vision/image_searcher_test.py#L105). Do you have any suggestions on how this needs to be handled?

khanhlvg: Yes there's an option to use the index file bundled in the TFLite file so the `if self.index_file_name is not None or self.index_file_content is not None:` check shouldn't be necessary. I'll remove it when importing this PR for internal review.

khanhlvg (approved):

tensorflow_lite_support/python/task/processor/proto/search_result_pb2.py
khanhlvg: After revising the code, It turned out that bytearray is the correct type because users may want to use non-string metadata to save space. Therefore, I reverted the commit.

khanhlvg (approved):

tensorflow_lite_support/python/task/processor/proto/search_result_pb2.py
kinaryml: @khanhlvg Yes, that was my observation too from the proto documentation so I kept it as a bytearray.

tensorflow_lite_support/python/task/processor/proto/search_options_pb2.py
khanhlvg: Ah now I got what you mean. Yes having the if check as you're doing now is correct.

PR #828, opened 2022-05-30
Task Library: NL Classifier Python Implementation

kinaryml:
- Added NLClassifier implementation files for the Python task API.


tensorflow_lite_support/cc/task/processor/proto/nl_classification_options.proto
```diff
@@ -0,0 +1,108 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+syntax = "proto2";
+
+package tflite.task.processor;
+
+option java_multiple_files = true;
```
khanhlvg, line 20: These two Java options are not being used so please remove them.

tensorflow_lite_support/python/task/processor/proto/nl_classification_options_pb2.py
```diff
@@ -0,0 +1,116 @@
...
+  Attributes:
+    input_tensor_name: Name of the input text tensor, if the model has
+      multiple inputs. Only the input tensor specified will be used for
+      inference; other input tensors will be ignored. Defaults to "INPUT".
+    output_score_tensor_name: Name of the output score tensor, if the model has
+      multiple outputs. Defaults to "OUTPUT_SCORE".
+    output_label_tensor_name: Name of the output label tensor, if the model has
+      multiple outputs. Defaults to "OUTPUT_LABEL". By default, label file
+      should be packed with the output score tensor through Model Metadata. See
+      the MetadataWriter for NLClassifier [1]. NLClassifier reads and parses
```
khanhlvg, line 57: docstring will be parsed as markdown so you can link to the MetadataWriter for NLClassifier site using Markdown syntax.

khanhlvg (approved):
Looks great. Thanks Kinar!

I left just a few minor comments. As they are minor changes, I'll fix them myself when putting this PR through internal review. You don't need to address them on GitHub.

tensorflow_lite_support/python/task/text/nl_classifier.py
```diff
@@ -0,0 +1,101 @@
...
+_CppNLClassifier = _pywrap_nl_classifier.NLClassifier
+_BaseOptions = base_options_module.BaseOptions
+_NLClassificationOptions = nl_classification_options_pb2.NLClassificationOptions
+
+
+@dataclasses.dataclass
+class NLClassifierOptions:
+  """Options for the NL classifier task."""
+  base_options: _BaseOptions
+  nl_classification_options: _NLClassificationOptions = _NLClassificationOptions()
```
khanhlvg, line 32: @kinaryml We got feedback from internal review that the Python NLClassifierOptions shouldn't expose these customization options that exist in C++ because it is meant to be for advanced users only. So please remove this proto and use the default options.

Instead, this should be replaced by the [ClassificationOptions](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/python/task/processor/proto/classification_options_pb2.py). These filtering options aren't available in the C++ layer so you should implement them in the Python layer.

khanhlvg (changes requested):

tensorflow_lite_support/python/task/text/nl_classifier.py
kinaryml: @khanhlvg I've added the filtering options in the Python layer along with some unit tests and removed any customization options associated with `NLClassificationOptions`. PTAL.

tensorflow_lite_support/cc/task/processor/proto/nl_classification_options.proto
```diff
@@ -17,9 +17,6 @@ syntax = "proto2";
 
 package tflite.task.processor;
 
-option java_multiple_files = true;
-option java_package = "org.tensorflow.lite.task.processor.proto";
-
 // Options for setting up an NLClassifier.
 // Next Id: 7
 message NLClassificationOptions {
```
khanhlvg, line 22: This file is no longer necessary because you don't need to pass these config from Python layer to pybinds.

tensorflow_lite_support/python/task/processor/proto/nl_classification_options_pb2.py
```diff
@@ -13,104 +13,6 @@
 # limitations under the License.
 """NL Classification options protobuf."""
```
khanhlvg, line 15: same here - this file is no longer neccessary.

tensorflow_lite_support/python/task/text/nl_classifier.py
```diff
@@ -71,12 +73,25 @@ def create_from_options(cls, options: NLClassifierOptions) -> "NLClassifier":
...
+
+    if classification_options.category_name_allowlist is not None and \
+        classification_options.category_name_denylist is not None:
+      if len(classification_options.category_name_allowlist) > 0 and \
+          len(classification_options.category_name_denylist) > 0:
+        raise ValueError(
+          "`class_name_allowlist` and `class_name_denylist` are mutually "
+          "exclusive options.")
+
     classifier = _CppNLClassifier.create_from_options(
```
khanhlvg, line 93: You should update the _CppNLClassifier to take only the BaseOptions.

tensorflow_lite_support/cc/task/processor/proto/nl_classification_options.proto
kinaryml: @khanhlvg Removed it.

tensorflow_lite_support/python/task/processor/proto/nl_classification_options_pb2.py
kinaryml: Removed it.

tensorflow_lite_support/python/task/text/nl_classifier.py
kinaryml: Updated.

khanhlvg (approved):

tensorflow_lite_support/python/task/text/nl_classifier.py
```diff
@@ -0,0 +1,163 @@
...
+    Raises:
+      ValueError: If any of the input arguments is invalid.
+      RuntimeError: If failed to perform NL classification.
+    """
+    classification_result = self._classifier.classify(text)
+    classification_result = classifications_pb2.ClassificationResult.\
+        create_from_pb2(classification_result)
+    return self._postprocess(classification_result)
+
+  def _postprocess(self, result: classifications_pb2.ClassificationResult):
```
lu-wang-g, line 113: Ideally, this logic should be checked into the C++ layer (in fact in C++, we can reuse the existing classification_postprocessor), and be shared across all platforms. My concern is the C++/Java NLClassifier don't support those options (i.e. allowlist/denylist) at this point, and this gonna be a little confusing why we only support it in Python. I'd be OK if we completely ignore this feature at this point. Otherwise, I would enable it from C++ and expose the options in other languages.

khanhlvg: SGTM. Let me update the internal code to ignore the ClassificationOptions entirely for now. We'll look into implementing it in the C++ layer later.

PR #830, opened 2022-06-03
iOS Image Embedder Objective C++

priankakariatyml:
1.  Added a test .mm file and corresponding test for Image Embedder
2. Added a new .sh file for building framework archive
3. Added new podspec for objective c++ frameworks


tensorflow_lite_support/ios/ios.bzl
```diff
@@ -28,3 +28,32 @@ def strip_c_api_include_path_prefix(name, hdr_labels, prefix = ""):
             > "$@"
             """.format(prefix, hdr_label),
         )
+
+# When the static framework is built with bazel, the all header files are moved
```
schmidt-sebastian, line 32: s/the all/all/

tensorflow_lite_support/ios/ios.bzl
```diff
@@ -28,3 +28,32 @@ def strip_c_api_include_path_prefix(name, hdr_labels, prefix = ""):
...
+# to the "Headers" directory with no header path prefixes. This auxiliary rule
+# is used for stripping the path prefix to the iOS API header files imported by
```
schmidt-sebastian, line 34:
```suggestion
# is used for stripping the path prefix from the iOS API header files imported by
```

tensorflow_lite_support/ios/TensorFlowLiteTaskVisionObjCPP.podspec
```diff
@@ -0,0 +1,20 @@
+Pod::Spec.new do |s|
+  s.name             = 'TensorFlowLiteTaskVisionObjCPP'
```
khanhlvg, line 2: Instead of having a dedicated podspec for the ObjC++ implementation, can you add the `TensorFlowLiteTaskVisionObjCPP.framework` as a vendored framework for the current TensorFlowLiteTaskVision pod? We don't want to expose the internal structure of the framework to the end-users if possible.

khanhlvg: Thanks Prianka! LGTM.

@schmidt-sebastian @lu-wang-g Do you have any feedback?

khanhlvg: @priankakariatyml Can you check if the new setup can make the Vision pod and Text pod co-exist in the same project? I guess that you may tweak the code to hide the duplicating symbols.

tensorflow_lite_support/ios/task/vision/sources/TFLImageEmbedder.mm
```diff
@@ -0,0 +1,50 @@
...
+
+    cc_options.mutable_model_file_with_metadata()->set_file_name(modelPath.UTF8String);
+
+    StatusOr<std::unique_ptr<ImageEmbedderCpp>> embedder_status =
+      ImageEmbedderCpp::CreateFromOptions(cc_options);
+ 
+    if (embedder_status.ok()) {
+        cppImageEmbedder = std::move(embedder_status.value());
+    }
+    else {
```
schmidt-sebastian, line 43:
```suggestion
    } else {
```

tensorflow_lite_support/ios/TensorFlowLiteTaskVision.podspec.template
```diff
@@ -1,62 +1,21 @@
...
-    'HEADER_SEARCH_PATHS' =>
-      '"${PODS_TARGET_SRCROOT}/' + root_dir_name + '/.."/**',
-    'VALID_ARCHS' => 'x86_64 armv7 arm64',
+     'VALID_ARCHS' => 'x86_64, arm64, armv7',
   }
-
   s.library = 'c++'
-  s.vendored_frameworks = 'Frameworks/TensorFlowLiteTaskVisionC.framework'
+  s.frameworks = 'CoreMedia', 'Accelerate'
+  s.vendored_frameworks = 'Frameworks/TensorFlowLiteTaskVisionObjCPP.framework'
```
yyoon, line 20: Shouldn't this be identical to the "bundle" name of the `ios_static_framework` rule?

tensorflow_lite_support/ios/task/vision/BUILD
```diff
@@ -72,3 +72,19 @@ objc_library(
...
+    ],
+    hdrs = [
+        "sources/TFLImageEmbedder.h",
+    ],
+    features = ["-layering_check"],
+    module_name = "TFLImageEmbedder",
+    deps = [
+        "//tensorflow_lite_support/cc/task/vision:image_embedder",
+    ],
+    copts = ["-ObjC++","-std=c++14"],
```
yyoon, line 89: Why is the `-ObjC++` option needed?

yyoon (commented):
Looks good to me at a high level.

As far as I can tell, this approach would work because the ObjC API headers (e.g., `TFLImageEmbedder.h`) are only using the other ObjC API headers in the same directory and not including any other C++ headers directly.

As noted, the header prefix stripping macro is probably not needed, but I'm not 100% sure about that. Please double check.

tensorflow_lite_support/ios/ios.bzl
```diff
@@ -28,3 +28,32 @@ def strip_c_api_include_path_prefix(name, hdr_labels, prefix = ""):
             > "$@"
             """.format(prefix, hdr_label),
         )
+
+# When the static framework is built with bazel, all the header files are moved
+# to the "Headers" directory with no header path prefixes. This auxiliary rule
+# is used for stripping the path prefix from the iOS API header files imported by
+# other iOS API header files.
+def strip_ios_api_include_path_prefix(name, hdr_labels, prefix = ""):
```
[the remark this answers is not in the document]
priankakariatyml: Yes stripping will be needed. Right now only one file has been added for reviewing the structure. In a full implementation the header would have a lot of imports in a hierarchical structure like [TFLImageClassifier.h](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/ios/task/vision/sources/TFLImageClassifier.h#L17). We will have to strip the prefixes and build the final iOS framework like the C framework is built.
I added a separate method since objective c files are imported using `#import` as opposed to c files which are included using `#include`. I can incorporate this check in  `strip_c_api_include_path_prefix` and call it `strip_api_include_path_prefix` which will strip path prefixes irrespective of  `#import ` or `#include`

tensorflow_lite_support/ios/TensorFlowLiteTaskVision.podspec.template
priankakariatyml: Sorry, my bad. While testing the pod I had named the bundle TensorFlowLiteTaskVisionObjCPP. Afterwards I changed it to TensorFlowLiteTaskVision but didn't change it here. I have updated it now.

tensorflow_lite_support/ios/task/vision/BUILD
priankakariatyml: My understanding was that "-ObjC++" is needed when you are building with `.mm` files. Or source file is `.mm`.

tensorflow_lite_support/ios/ios.bzl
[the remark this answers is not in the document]
yyoon: I see. If it's needed, I'd recommend consolidating these two functions into a single one as mentioned before.

PR #837, opened 2022-06-13
Task Library: BertNLClassifier Python implementation

kinaryml:
- Python API implementation for the BertNLClassifier
- A simple test to try out the API


tensorflow_lite_support/python/task/text/bert_nl_classifier.py
```diff
@@ -0,0 +1,102 @@
...
+from tensorflow_lite_support.python.task.processor.proto import classifications_pb2
+from tensorflow_lite_support.python.task.text.pybinds import _pywrap_bert_nl_classifier
+
+_CppBertNLClassifier = _pywrap_bert_nl_classifier.BertNLClassifier
+_BaseOptions = base_options_module.BaseOptions
+
+
+@dataclasses.dataclass
+class BertNLClassifierOptions:
+  """Options for the Bert NL classifier task."""
```
khanhlvg, line 29: Please add docstring to explain each parameter. They will show up in API documentation like this:
https://www.tensorflow.org/lite/api_docs/python/tflite_support/task/audio/AudioClassifierOptions

I realized that some option classes don't have parameter docstring so please send a follow-up PR to fix other classes.

khanhlvg (changes requested):

kinaryml:
@khanhlvg I've raised a small PR to address the parameter docstrings for the other tasks here.

https://github.com/tensorflow/tflite-support/pull/838

tensorflow_lite_support/python/task/text/bert_nl_classifier.py
```diff
@@ -0,0 +1,107 @@
...
+@dataclasses.dataclass
+class BertNLClassifierOptions:
+  """Options for the Bert NL classifier task.
+
+  Attributes:
+    base_options: Base options for the Bert NL classifier task.
+    max_seq_len: Max number of tokens to pass to the model.
+  """
+  base_options: _BaseOptions
+  max_seq_len: Optional[int] = 128
```
khanhlvg, line 36: Actually it looks like max_seq_len is now deprecated so you can remove this option.
https://github.com/tensorflow/tflite-support/blob/46cd03695e156ff67141b73eb44579e412f3e44d/tensorflow_lite_support/cc/task/text/proto/bert_nl_classifier_options.proto#L33

kinaryml: @khanhlvg Alright, I've removed it now.

khanhlvg (approved):

PR #850, opened 2022-07-07
Fixed On Device Memory Issue

priankakariatyml:
Fixed an EXC_BAD_ACCESS error which occurred on device due to early release of `audioTensor.buffer.data` before the end of 
`-[TFLAudioClassifier classifyWithAudioTensor:error:]` if it is not assigned to a local variable in the method.

### DIFF
--- a/tensorflow_lite_support/ios/task/audio/sources/TFLAudioClassifier.m
+++ b/tensorflow_lite_support/ios/task/audio/sources/TFLAudioClassifier.m
@@ -155,7 +155,6 @@
   TfLiteAudioBuffer cAudioBuffer = [audioTensor cAudioBufferFromFloatBuffer:audioTensorBuffer];
 
   TfLiteSupportError *classifyError = NULL;
-
   TfLiteClassificationResult *cClassificationResult =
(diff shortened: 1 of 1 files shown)


khanhlvg (approved):

tensorflow_lite_support/ios/task/audio/sources/TFLAudioClassifier.m
```diff
@@ -151,7 +151,8 @@ - (nullable TFLClassificationResult *)classifyWithAudioTensor:(TFLAudioTensor *)
     return nil;
   }
 
-  TfLiteAudioBuffer cAudioBuffer = [audioTensor cAudioBufferFromFloatBuffer:audioTensor.buffer];
+  TFLFloatBuffer *audioTensorBuffer = audioTensor.buffer;
+  TfLiteAudioBuffer cAudioBuffer = [audioTensor cAudioBufferFromFloatBuffer:audioTensorBuffer];
```
khanhlvg, line 155: This is an interesting behavior. `audioTensor.buffer` actually creates a new TFLFloatBuffer object and if it isn't stored as a local variable, it can be released before the inference code has finished executing.

Thanks for identifying and fixing the issue.

PR #851, opened 2022-07-08
Task Library: Added an `offset` param to the `LoadAudioBufferFromFile` C++ function for flexibility

kinaryml:
- Updated some C tests
- Added an offset param to `create_from_wav_file` in the Python Task API

### DIFF
--- a/tensorflow_lite_support/cc/task/audio/utils/audio_utils.cc
+++ b/tensorflow_lite_support/cc/task/audio/utils/audio_utils.cc
@@ -20,7 +20,7 @@ namespace task {
 namespace audio {
 
 tflite::support::StatusOr<AudioBuffer> LoadAudioBufferFromFile(
-    const std::string& wav_file_path, int buffer_size, uint32_t* offset,
+    const std::string& wav_file_path, uint32_t* buffer_size, uint32_t* offset,
     std::vector<float>* wav_data) {
   std::string contents = ReadFile(wav_file_path);
 
@@ -32,8 +32,8 @@ tflite::support::StatusOr<AudioBuffer> LoadAudioBufferFromFile(
       contents, wav_data, offset, &decoded_sample_count, &decoded_channel_count,
       &decoded_sample_rate));
 
-  if (decoded_sample_count > buffer_size) {
-    decoded_sample_count = buffer_size;
+  if (decoded_sample_count > *buffer_size) {
+    decoded_sample_count = *buffer_size;
   }
 
   return AudioBuffer(
--- a/tensorflow_lite_support/cc/task/audio/utils/audio_utils.cc
+++ b/tensorflow_lite_support/cc/task/audio/utils/audio_utils.cc
@@ -20,7 +20,7 @@ namespace task {
 namespace audio {
 
 tflite::support::StatusOr<AudioBuffer> LoadAudioBufferFromFile(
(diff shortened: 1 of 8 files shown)


tensorflow_lite_support/cc/task/audio/utils/wav_io.cc
```diff
@@ -113,10 +113,10 @@ absl::Status ReadString(const std::string& data, int expected_length,
 
 absl::Status DecodeLin16WaveAsFloatVector(const std::string& wav_string,
                                           std::vector<float>* float_values,
+                                          int offset,
```
khanhlvg, line 116: Please use `uint32_t` to align with the type of sample_count

khanhlvg (approved):

tensorflow_lite_support/cc/task/audio/utils/wav_io.cc
kinaryml: The original data type for the offset was `int` though.

https://github.com/tensorflow/tflite-support/blob/72b4ca66479054d0463a0e52b59e7a7b2da8ecc0/tensorflow_lite_support/cc/task/audio/utils/wav_io.cc#L119

khanhlvg: Could you change it to uint32_t? It's better than int because offset can't be negative.

I think it'll require you to change the datatype for offset of underlying methods but I think it makes more sense to align the datatype of offset with that of sample_count through out the codebase.

kinaryml: Alrighty, I'll make those changes.

kinaryml:
@khanhlvg I've made some changes now. PTAL and let me know if it looks good.

tensorflow_lite_support/cc/task/audio/utils/audio_utils.cc
```diff
@@ -20,7 +20,7 @@ namespace task {
 namespace audio {
 
 tflite::support::StatusOr<AudioBuffer> LoadAudioBufferFromFile(
-    const std::string& wav_file_path, int buffer_size, int offset,
+    const std::string& wav_file_path, int buffer_size, uint32_t* offset,
```
khanhlvg, line 23: Let's change the buffer_size here to uint32_t as well to match.

tensorflow_lite_support/cc/task/audio/utils/wav_io.cc
```diff
@@ -62,17 +62,17 @@ std::string ReadFile(const std::string filepath) {
 
 // Handles moving the data index forward, validating the arguments, and avoiding
 // overflow or underflow.
-absl::Status IncrementOffset(int old_offset, size_t increment, size_t max_size,
+absl::Status IncrementOffset(uint32_t *old_offset, size_t increment, size_t max_size,
```
khanhlvg, line 65: Shouldn't new_offset be uint32_t here as well to match data type?

kinaryml:
@khanhlvg I've fixed and addressed some of the feedback you've left FYI. Oh and I've also tested the 3 audio classifier files (C/C++/Python  demo + test code) to ensure it runs fine now.

khanhlvg (approved):
Thanks Kinar!

PR #853, opened 2022-07-13
Added iOS Audio Classifier podspec

priankakariatyml:


tensorflow_lite_support/ios/BUILD
```diff
@@ -179,3 +184,24 @@ ios_static_framework(
...
+        ":TFLCategory.h",
+        ":TFLClassificationResult.h",
+        ":TFLAudioFormat.h",
+        ":TFLFloatBuffer.h",
+        ":TFLAudioRecord.h",
+        ":TFLAudioTensor.h",
+        ":TFLAudioClassifier.h",
+    ],
+    bundle_name = "TensorFlowLiteTaskAudio",
+    minimum_os_version = "10.0",
```
khanhlvg, line 203: Can this be 9.0 (TFL_MINIMUM_OS_VERSION) rather than 10.0?

If it needs to be 10.0, could you define this as a constant (e.g. TFL_TASK_MINIMUM_OS_VERSION) to share with the vision pod?

khanhlvg (changes requested):

tensorflow_lite_support/ios/BUILD
priankakariatyml: According to our discussion, I have made the deployment target for audio consistent with vision (10.0) and retained text at 9.0

khanhlvg (approved):

PR #857, opened 2022-07-14
[NOT FOR MERGING] Core ML Delegate Tests on iOS Device

priankakariat:


tensorflow_lite_support/c/task/vision/image_classifier.cc
```diff
@@ -88,6 +88,17 @@ StatusOr<ImageClassifierOptionsCpp> CreateImageClassifierCppOptionsFromCOptions(
   cpp_options.set_score_threshold(
       c_options->classification_options.score_threshold);
 
+  cpp_options.mutable_base_options()
```
priankakariatyml, line 91: Hardcoding core ml delegate settings here since these changes are only for testing on a device.

tensorflow_lite_support/ios/task/vision/BUILD
```diff
@@ -24,6 +24,7 @@ objc_library(
         "//tensorflow_lite_support/ios/task/processor:TFLClassificationResultHelpers",
         "//tensorflow_lite_support/ios/task/vision/utils:GMLImageUtils",
         "//tensorflow_lite_support/odml/ios/image:MLImage",
+        "@org_tensorflow//tensorflow/lite/experimental/acceleration/configuration:coreml_plugin",
```
priankakariatyml, line 27: I have linked in the cormel_plugin target here. I am still getting the following error:
"Could not create CoreML plugin. Have you linked in the CoreML_plugin target?"

tensorflow_lite_support/c/task/vision/image_classifier.cc
priankakariatyml: Once issues are fixed these can be implemented in the iOS framework.

PR #862, opened 2022-07-19
Task Library: Added more tests for the BertCluAnnotator Python API

kinaryml:
- This also includes the BertCluAnnotator to the pip package.

Code snippet for using the BertCluAnnotator API:

```python
from tflite_support.task import text
from tflite_support.task import core
from tflite_support.task import processor

base_options = core.BaseOptions(file_name=model_path)
options = text.BertCluAnnotatorOptions(base_options=base_options)
annotator = text.BertCluAnnotator.create_from_options(options)
# annotator = text.BertCluAnnotator.create_from_file(model_path)
clu_request = processor.CluRequest(...)
text_clu_response = annotator.annotate(clu_request)
print(text_clu_response)
```

### DIFF
--- a/tensorflow_lite_support/python/test/task/text/bert_clu_annotator_test.py
+++ b/tensorflow_lite_support/python/test/task/text/bert_clu_annotator_test.py
@@ -39,60 +39,37 @@ _BertCluAnnotationOptions = clu_annotation_options_pb2.BertCluAnnotationOptions
 _BERT_MODEL = 'bert_clu_annotator_with_metadata.tflite'
 
 _CLU_REQUEST = _CluRequest(utterances=[
-  'Hi, I would like to make a dinner reservation.'
-  'Of course, what evening will you be joining us on?'
-  'We will need the reservation for Tuesday night.'
-  'What time would you like the reservation for?'
-  'We would prefer 7:00 or 7:30.'
-  'How many people will you need the reservation for?'
-  'There will be 4 of us.'
-  'Fine, I can seat you at 7:00 on Tuesday, if you would kindly give me your '
-  'name.'
-  'Thank you. The last name is Foster.'
-  'See you at 7:00 this Tuesday, Mr. Foster.'
-  'Thank you so much. I appreciate your help.'
+  'I would like to book a reservation at your hotel',
+  'What date would you like to make that reservation for?',
+  'I need the reservation for the 14th of May',
+  'How many days do you need the reservation for?',
+  'I will be staying for 3 nights.',
+  'Is that a single room, or will there be more guests?',
+  'I need a double room.',
+  'We have smoking and nonsmoking rooms. Which do you prefer?',
+  'We require a nonsmoking room.',
+  'Your room is booked. You must arrive before 4:00 pm the day you are to '
+  'check in.'
 ])
 _CLU_RESPONSE = _CluResponse(
   domains=[
     _Category(
-      index=0, score=0.925067, display_name='Restaurants',
-      category_name='')],
+      index=0, score=0.9158045053482056, display_name='Hotels',
+      category_name='')
+  ],
   intents=[
     _Category(
-      index=0, score=0.945974, display_name='ReserveRestaurant',
-      category_name='')],
-  categorical_slots=[
-    _CategoricalSlot(
-      slot='number_of_seats',
-      prediction=_Category(
-        index=0, score=0.784716, display_name='4', category_name=''))
+      index=0, score=0.6116464734077454, display_name='NOTIFY_SUCCESS',
+      category_name='')
   ],
+  categorical_slots=[],
   noncategorical_slots=[
     _NonCategoricalSlot(
-      slot='restaurant_name', extraction=_Extraction(
-        value='Tuesday night', score=0.603665, start=129, end=142)),
-    _NonCategoricalSlot(
-      slot='time', extraction=_Extraction(
-        value='7:00', score=0.907133, start=204, end=208)),
-    _NonCategoricalSlot(
-      slot='time', extraction=_Extraction(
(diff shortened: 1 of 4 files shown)


tensorflow_lite_support/python/test/task/text/bert_clu_annotator_test.py
```diff
@@ -131,6 +133,73 @@ def test_annotate_model(self, model_name, model_file_type, clu_request,
...
+    # Annotates CLU request using the given model.
+    text_clu_response = annotator.annotate(clu_request)
+    self.assertProtoEquals(text_clu_response.to_pb2(),
+                           expected_clu_response.to_pb2())
+
+  def test_max_history_turns(self):
+    # Creates annotator.
+    base_options = _BaseOptions(file_name=self.model_path)
+    bert_clu_annotation_options = _BertCluAnnotationOptions(
+        max_history_turns=10)
```
khanhlvg, line 171: The input only has 3 exchanges so setting max_history_turns=10 doesn't change the model output.

Could you either add more turns (sentences) to the conversation and adjust max_history_turns so that it doesn't cover the whole conversation? Ideally, I think you should change both _CLU_REQUEST and _CLU_RESPONSE, making the conversation longer, and here set a max_history_turns shorter than the conversation length to test that the output does change. Please note that the default value of max_history_turns is 5.

khanhlvg (changes requested):

tensorflow_lite_support/python/test/task/text/bert_clu_annotator_test.py
```diff
@@ -37,33 +37,62 @@
 _BertCluAnnotationOptions = clu_annotation_options_pb2.BertCluAnnotationOptions
 
 _BERT_MODEL = 'bert_clu_annotator_with_metadata.tflite'
+
 _CLU_REQUEST = _CluRequest(utterances=[
-    'I would like to make a restaurant reservation at morning 11:15.',
-    'Which restaurant do you want to go to?',
-    'Can I get a reservation for two people at Andes Cafe? Where is their ' +
-    'address?'
+  'Hi, I would like to make a dinner reservation.'
```
khanhlvg, line 42: You're using one long sentence. Please split them into sentences where speakers change. Without it, the max_history_turns won't have any effect because it see everything as one turn.

For example:
```
'Hi, I would like to make a dinner reservation.',
'Of course, what evening will you be joining us on?'
```

kinaryml: @khanhlvg Oh sorry about that. I've fixed it now. PTAL again.

khanhlvg (approved):

tensorflow_lite_support/python/test/task/text/bert_clu_annotator_test.py
khanhlvg: Thanks Kinar for the quick fix.

As a sanity check, could you test directly using the C++ API and see whether the same list of utterances with the default config gives the same result with the C++ API? From the expected response, I see that the model wasn't able to pick up info such as reservation dates, room type etc., which are important for a hotel reservation. It can well be the limitation of the model itself, but it's safer to check and make sure that it's not a bug in the Python implementation.

kinaryml: @khanhlvg The C++ API seems to generate the same results from my testing. So I believe it has to do with the model itself not being able to catch all of the info. You can feel free to try it out yourself too. I just happened to replace a [test case](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/cc/test/task/text/bert_clu_annotator_test.cc#L129-L157) with the following lines of code and it should give the same results as you can see in the Python API:

```cpp
TEST(BertCluAnnotatorTest, TestAnnotatorBasic) {
  BertCluAnnotatorOptions options;
  options.mutable_base_options()->mutable_model_file()->set_file_name(
      GetFullPath(kTestBertCluAnnotatorModelWithMetadataPath));
  SUPPORT_ASSERT_OK_AND_ASSIGN(std::unique_ptr<CluAnnotator> clu_annotator,
                       BertCluAnnotator::CreateFromOptions(options));
  CluRequest request;
  ASSERT_TRUE(TextFormat::ParseFromString(
      R"pb(
            utterances: "I would like to book a reservation at your hotel"
            utterances: "What date would you like to make that reservation for?"
            utterances: "I need the reservation for the 14th of May"
            utterances: "How many days do you need the reservation for?"
            utterances: "I will be staying for 3 nights."
            utterances: "Is that a single room, or will there be more guests?"
            utterances: "I need a double room."
            utterances: "We have smoking and nonsmoking rooms. Which do you prefer?"
            utterances: "We require a nonsmoking room."
            utterances: "Your room is booked. You must arrive before 4:00 pm the day you are to check in."
      )pb",
      &request));
  SUPPORT_ASSERT_OK_AND_ASSIGN(const auto response, clu_annotator->Annotate(request));

  ExpectApproximatelyEqual(
      response, ParseTextProtoOrDie<CluResponse>(R"pb(
            domains {
              index: 0
              score: 0.915805
              display_name: "Hotels"
              class_name: ""
            }
            intents {
              index: 0
              score: 0.611646
              display_name: "NOTIFY_SUCCESS"
              class_name: ""
            }
            noncategorical_slots {
              slot: "time"
              extraction {
                value: "4:00 pm"
                score: 0.794008
                start: 44
                end: 51
              }
            }
      )pb"));
}
```

khanhlvg: Thanks Kinar for verifying. We're still waiting for internal review so I'll keep you updated on how it goes.

PR #893, opened 2022-10-20
Added Swift Tests for iOS Image Searcher and include ImageSearcher API in TFLiteTaskVision pod

priankakariatyml:
1. Added Swift Tests
2. Included image searcher in iOS vision framework for coocoapods

### DIFF
--- a/tensorflow_lite_support/ios/BUILD
+++ b/tensorflow_lite_support/ios/BUILD
@@ -162,6 +162,10 @@ objc_library(
         "@com_google_absl//absl/strings",
         "@com_google_absl//absl/strings:cord",
     ],
+     copts = [
+        "-ObjC++",
+        "-std=c++17",
+    ],
 )
 
 # bazel build -c opt --config=ios_arm64 //tensorflow_lite_support/ios:TensorFlowLiteTaskVision_framework
--- a/tensorflow_lite_support/ios/BUILD
+++ b/tensorflow_lite_support/ios/BUILD
@@ -44,15 +44,20 @@ strip_api_include_path_prefix(
         "//tensorflow_lite_support/ios/task/text/qa:Sources/TFLBertQuestionAnswerer.h",
         "//tensorflow_lite_support/ios/task/vision:sources/TFLImageClassifier.h",
         "//tensorflow_lite_support/ios/task/vision:sources/TFLObjectDetector.h",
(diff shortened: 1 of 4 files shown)


priankakariatyml:
@khanhlvg @schmidt-sebastian Raised a PR for Swift Tests

tensorflow_lite_support/ios/BUILD
```diff
@@ -213,6 +229,7 @@ ios_static_framework(
         ":TFLAudioRecord.h",
         ":TFLAudioTensor.h",
         ":TFLBaseOptions.h",
+        ":TFLExternalFile.h",
```
khanhlvg, line 232: Why do you need to add TFLExternalFile to the Audio framework?

priankakariatyml: [TFLBaseOptions](https://github.com/tensorflow/tflite-support/blob/master/tensorflow_lite_support/ios/task/core/sources/TFLBaseOptions.h#L84) also needs `TFLExternalFile` for defining the model file. Before the image searcher APIs were added `TFLExternalFile` was defined in `TFLBaseOptions.h`.
But searcher needed `TFLExternalFile` in `TFLSearchOptions` as well for defining the index file. Hence the class definition  was moved to a separate file. So all frameworks that use `TFLBaseOptions` needs to include `TFLExternalFile.h` as well.

khanhlvg: I see. Thanks for tha clarification.

tensorflow_lite_support/ios/test/task/vision/image_searcher/TFLImageSearcherTests.swift
```diff
@@ -0,0 +1,178 @@
...
+        mlImage: mlImage))
+    self.validateSearchResult(searchResult)
+  }
+
+  func testCreateImageSearcherWithNoModelPathFails() throws {
+    let options = ImageSearcherOptions()
+    do {
+      let imageSearcher = try ImageSearcher.searcher(
+        options: options)
+      XCTAssertNil(imageSearcher)
```
khanhlvg, line 170: An error is already thrown when you create the imageSearcher so this assert will not get called. You'll need to do the assert outside of the do-catch statement.

I'll fix it when I import the PR.

khanhlvg (approved):
Just a minor issue that I co

PR #896, opened 2022-10-26
Fixed memory leak in iOS Image Searcher

priankakariatyml:
Fixed memory leak in allocation of image pixel buffer.


priankakariatyml:
@khanhlvg @schmidt-sebastian This PR fixes a memory leak in image searcher. The underlying pixel buffer created with malloc was not being deallocated leading to an out of memory issue in the camera app.

khanhlvg (approved):

PR #898, opened 2022-11-02
Added iOS Text Searcher and Basic Objective C Test

priankakariatyml:


priankakariatyml:
@khanhlvg @schmidt-sebastian I have raised a PR for text searcher and one basic Obj C test

PR #916, opened 2023-03-04
Fixed typos and added additional example for usage

anjakuchenbecker:
1) Fixed several typos. 
2) Added additional usage example besides the existing "Transferring metadata and associated files from another TFLite model", as this describes only the model buffer approach. From my point of view an additional model file approach example is useful here, as this is the needed after you compiled your tflite model for an EdgeTPU (with EdgeTPU compiler) to transfer the metadata, including the label file, from the original tflite model to the EdgeTPU model.

### DIFF
--- a/tensorflow_lite_support/metadata/python/metadata.py
+++ b/tensorflow_lite_support/metadata/python/metadata.py
@@ -109,16 +109,16 @@ class MetadataPopulator(object):
   third_party/tensorflow/lite/schema/metadata_schema.fbs
 
   Example usage:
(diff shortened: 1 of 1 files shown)


anjakuchenbecker:
Hi @lu-wang-g 

asking here for code review as I'm not allowed to assign you as reviewer in the reviewers sidebar. 

Thank you
Anja

lu-wang-g (approved):

lu-wang-g:
Thanks for creating the fix, @anjakuchenbecker !

PR #1013, opened 2026-07-24
jni_utils: IWYU for recent libc++

nickdesaulniers:
libc++ is getting stricter about its transitive includes.  IWYU to avoid errors around malloc being undefined.


nickdesaulniers:
cc @yikong

lu-wang-g (approved):

